PR #4, opened 2018-01-12
Feature/multichannel

pslocum:


pslocum:
Deleting without merging.

PR #5, opened 2018-01-12
Updated to Kassiopeia 3.3.2

pslocum:
Updated locust to be compatible with latest kassiopeia upgrade.
KToolbox was moved from Kassiopeia to Kommon


pslocum:
deleted branch.

PR #20, opened 2018-05-08
Develop

pslocum:
More streamlined selection of Phases 1-3 from the LMCCyclotronRadiationExtractor in the xml file.


pslocum:
Streamlining phase 1-3 selection from the xml file.

PR #21, opened 2018-06-07
Feature/buzinsky patches

buzinsky:
Class Structure which should be more modular to different types of receivers


buzinsky:
That sounds correct. If we want additional antenna designs worked into locust, they would be new classes of the same form as LMCPatchAntenna. The classes (LMCChannel, LMCReceiver, LMCPatchAntenna) are really the ones I care about, which should make it easier/ more modular to configure antenna configs

pslocum:
Merged with develop.

PR #31, opened 2018-10-01
Redid FindNode(), pileup code

buzinsky:
May want to update FindNode() in other generators ( I forget if they use them)


pslocum:
Merged Nick's new FindNode() and pileup calculations.

PR #42, opened 2018-12-12
implemented jump size following scattering pdf from Assev, fixed numb…

les67:
- Implemented event jump size following the scattering PDF from Aseev et al
- Fixed a bug regarding the number of total tracks in an event


pslocum:
merged.

PR #46, opened 2019-01-15
ceil instead of round ntracks_val

cclaessens:
The number of tracks per event in the data follows an exponential distribution.
So far the number of tracks per event in the fake track generator was calculated by rounding a sample from an exponential distribution. To avoid simulations with zero tracks per event, the number of tracks is set to 1 if the rounded sample is zero. 

But this results in a massive distortion of the ntracks per event distribution after reconstruction. To fix this it was suggested to always add 1 after rounding and therefore shift the entire distribution up. However, that would not complete fix the issue, because the range that is rounded to 1 is twice as large as the range that is rounded to zero.
I suggest that instead of rounding we take the ceil of the random sample. This way, the distribution is still shifted up, there wouldn't be any events with zero tracks and the problem of non-equal rounding ranges is avoided.

Question:
The configurable parameter 'ntracks-mean' is not actually the mean track length (neither before nor after this pull request). Should we rename it?

### DIFF
--- a/Source/Generators/LMCFakeTrackSignalGenerator.cc
+++ b/Source/Generators/LMCFakeTrackSignalGenerator.cc
@@ -453,7 +453,7 @@ namespace locust
 
         std::default_random_engine generator(random_seed_val);
         std::exponential_distribution<double> ntracks_distribution(1./fNTracksMean);
-        ntracks_val = round(ntracks_distribution(generator));
+        ntracks_val = ceil(ntracks_distribution(generator));
         if ( ntracks_val == 0 ) // if we rounded to 0, let's simulate at least one tracks
         {
             ntracks_val = 1;


pslocum:
We could name ntracks_mean to ntracks_avg.  Also, did we still want to keep these lines below, or should they be deleted if we are using "ceil" instead of "round"?

if ( ntracks_val == 0 ) // if we rounded to 0, let's simulate at least one tracks
        {
            ntracks_val = 1;
        }

les67:
"ntracks_mean" is not the mean track length, it is the mean of the number of tracks in a given event. The mean track length is set by the parameter "track-length-mean". Thus I don't support the name change.

As far as the ceiling solution goes, it is a good one.

cclaessens:
sorry, I meant to say mean number of tracks not mean track length.
I don't know a better way to describe this parameter though. It will be the mean of the number of tracks if the the events with zero tracks were included. Maybe it's just fine if we leave it this way?

Can the sample ever be exactly zero, and therefore the ceil still be 0 too? If note, the lines below (if ntracks_val == 0 aso) could be removed.

les67:
I would say leave the parameter name that way lest we make it a more descriptive but longer and cumbersome name. At the same time, it would be important to note it in the Locust README for the fake event generator and explain what it actually represents. I would update this in my tutorial slides too. But I'm open to changing the name if both of you think of a better one. 

Good question, I would bet that we would never get exactly zero but perhaps we should leave the if statement in place just in case; perhaps add a comment on that line explaining why it's there? Something like "# In case we ever get exactly 0".

pslocum:
merged.

PR #53, opened 2019-02-15
Feature/multiple records per acq

cclaessens:
At the moment, every record is written to a new acquisition.
With these changes, the records in egg files will contain a single acquisition with n-records.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ cmake_minimum_required( VERSION 3.1 )
 
 # Define the project
 cmake_policy( SET CMP0048 NEW ) # version in project()
-project( locust_mc VERSION 1.9.5)
+project( locust_mc VERSION 1.9.6)
 
 list( APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/Scarab/cmake )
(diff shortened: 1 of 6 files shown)


cclaessens:
I just checked the LMCFakeTrackSignalGenerator and in DoGenerateTime the time is always counted up starting from zero. The problem I see now, is that because the LMCSimulationController creates a new signal for every record and then DoGenerateTime is executed, the FakeTrackSignalGenerator will never get past the sample_rate * record_size in time.

I think what we want to do is that the LMCSimulationController just creates one signal for every simulation, which is then processed by all generators and only split into records after being digitized.

pslocum:
image
I found the problem that was causing the high noise values.  The fAcquisitionRate is initialized at a different time than the LMCGaussianNoiseGenerator, and was being applied to the noise sigma every time the LMCGaussianNoiseGenerator was run (at each new record).  The multiple records look improved now (possible image above).  I am going to merge this feature branch into develop but am not going to delete it.

pslocum:
Combine multiple records into one acquisition to model streaming data.  Bug fix in how fAcquisitionRate is applied in multi-record LMCGaussianNoiseGenerator simulation.

PR #54, opened 2019-02-18
Using scarab v1.6.4 and changed fADCValuesSigned to default true in L…

cclaessens:
…MCDigitizer

So that singed ints intstead of unsigned are written to egg files.

### DIFF
--- a/Source/Generators/LMCDigitizer.cc
+++ b/Source/Generators/LMCDigitizer.cc
@@ -22,7 +22,7 @@ namespace locust
 
     Digitizer::Digitizer( const std::string& aName ) :
             Generator( aName ),
-            fADCValuesSigned( false )
+            fADCValuesSigned( true )
     {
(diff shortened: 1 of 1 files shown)


pslocum:
This looks good.  I want to double-check the multi-channel functionality in Locust, probably on Wednesday, and then we should merge this.

cclaessens:
There is a problem with this.
The samples in the egg files look fine, but the gain variation after katydid is distorted and now events are found. I am suspecting that it has to do with v-offset but I am not sure. Currently investigating...

pslocum:
I tested this branch using the LMCFakeTrackGenerator with Scarab 1.6.5 and noticed an artificial track above the LPF.  So I reverted all Locust branches to Scarab 1.6.4 for right now.

pslocum:
Closing this request without merging as Scarab reference is updated to v1.6.5 and fADCValuesSigned is now set to true on develop and master.

PR #68, opened 2019-06-13
Feature/plane wave calibration

arinabykadorova:
Added buffers and FIR filter to plane wave generator.


arinabykadorova:
Yes. I thought it was only plane wave generator files. What should I change?

PR #76, opened 2019-08-09
Feature/fake scatter

### LINKED ISSUE #81: Inconsistent variable naming and initialization
LMCFakeTrackGenerator.hh contains this block of member variable definitions:

```
            double slope_val = 0.;
            double pitch_val = 0.;
            double tracklength_val = 0.;
            double starttime_val = 0.;
            double endtime_val = 0.;
            double startfreq_val = 0.;
            double jumpsize_val = 0.002e9;
            int ntracks_val = 0;
```

The names should be consistent with the naming convention in Locust, which is to prepend with an `f` and use CamelCase.

Initialization should also be consistent with variable initialization elsewhere, which is done in the constructor(s).

buzinsky:
New Features:
* Scattering now has both energy losses and pitch angle changes
* Can scatter off of either H_2 or Kr
* Accounts for frequency changes from pitch angle changes (average B field)

Before merging:
- CMake needs to be sorted out (only runs in data files are in directory LocustSim is called from)
- Potential variable renaming/ cleanup

Edit: Aug. 21 / 19 (11 AM EST) (hopefully this works)
Resolves #81


buzinsky:
It looks like it ... when I look at https://www.boost.org/doc/libs/1_66_0/libs/math/doc/html/interpolation.html, I see the page exists for 1.66, 1.67 ... but not 1.65. So I think 1.66 is the minimum version we need

buzinsky:
Yes, this is on a mac. I am on 1.70. Clearly you have the header file, its just not liking the types of my arguments

buzinsky:
I see what the issue is: boost 1.70 is the first version that defines the constructor: 
`barycentric_rational(std::vector<Real>&& x, std::vector<Real>&& y, size_t order = 3);` 

See the links: https://www.boost.org/doc/libs/1_70_0/libs/math/doc/html/math_toolkit/barycentric.html vs https://www.boost.org/doc/libs/1_69_0/libs/math/doc/html/math_toolkit/barycentric.html

Either the min version should be boost 1.70, if using this (slightly) more convenient constructor, or 1.66 if we use c++ arrays [] in the constructor. I will make the change to the constructor so we are using the arrays, not the vectors. Its an extra 2 or 3 lines only

pslocum:
This does compile and run with Boost 1.66, if the data files are in the local directory.  It looks ready to be merged with develop, except for the data subdirectory to be built with Scarab v2.  Let us know if it needs to be merged without the new data subdirectory, or whether it should be merged as-is.

buzinsky:
Thanks. Lets wait until the Cmake is changed before merging. I do want to
have the Data directory in there, so people don't have to do this manually.
Noah put in the prerequisite changes to scarab late on Fri, so the
outstanding changes are just to the locust CMake. I will talk to you monday
if its either done, or I need help

pslocum:
OK, holding off on the merge and edited the README to reflect updated requirement to Boost 1.66.

pslocum:
Do we want to test this branch with the needed "data" directory before merging?

buzinsky:
> Do we want to test this branch with the needed "data" directory before merging?
Ya, I will rebase and do a final lookover in the morning (EST). I will comment when I am done w/ it/ it is ready to merge

pslocum:
What about if we merge develop into feature/fakeScatter and test in place, with all of the Scarab updates?

nsoblath:
I modified LMCFakeTrackGenerator so that it uses the data directory from PackageBuilder.  I tested it with a small test program, and similar code picked up the directory correctly.

I also moved the `Data` directory out of `Source/Core` to be its own top-level directory.

PR #78, opened 2019-08-16
Feature/scarabv2tests

buzinsky:
Okay, this compiles now: I think I did all the changes.
Need to check before merging (which I havent gotten to):
 - Does it compile with kassbuild?
 - Does an example generator run?
 - Are there any spelling errors (in get config entries) where compiler wont detect typos?
 - Any other mistakes

I will be on the SC call / lunch until 1:30 before I can get to this

### DIFF
--- a/Source/Applications/LocustSim.cc
+++ b/Source/Applications/LocustSim.cc
@@ -35,7 +35,7 @@ int main( int argc, char** argv )
 
         LPROG( lmclog, "Setting up generator toolbox" );
         GeneratorToolbox toolbox;
-        if( ! toolbox.Configure( &configurator.config() ) )
+        if( ! toolbox.Configure( configurator.config() ) )
         {
             LERROR( lmclog, "Unable to configure the generator toolbox" );
             return -1;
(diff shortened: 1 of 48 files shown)


pslocum:
I am compiling with Kass now and will run a short test when it finishes.

pslocum:
This compiles with kassiopeia.  Compiling without Kassiopeia throws a "map::at" exception for some generators ("decimate-signal") but not others ("gaussian-noise", "test-signal", "digitizer").  Will need to double check syntax in non-working generators.

buzinsky:
Great for first part. Interesting for second. I am able to compile w/o kassiopeia on my mac.

buzinsky:
Ive double checked all the changes, I dont see any typos by eye, so it is clear on my end (though you/ your machine may be picking something I am not)

pslocum:
I found the problem - in Scarab v2 the generators have to be fully implemented in the json file whether they have parameters or not.  In v1 we could just list the generators without needing an additional empty parameter array.  So I will test a few more things with this info.

nsoblath:
I just made a few updates in GeneratorToolbox.  Now it's once again optional to have a configuration block for a given generator.  I also left a few notes of things that I think should be changed, labeled as `TODO` in code comments.  If they look like reasonable changes, I can go ahead and make them.

pslocum:
I tested the update to LMCGeneratorToolbox, and if I omit the generator configuration block in the json file, there are alternating (infintely?) ongoing messages like this:
2019-08-16 21:57:21 [ INFO] (tid 139763379227456) eratorToolbox.cc(97): Configuring generator <lpf-fft>
2019-08-16 21:57:21 [DEBUG] (tid 139763379227456) eratorToolbox.cc(100): No configuration information present.  
In "release" instead of "debug", the code appears to hang.  In the meantime the map::at exception is now gone.

The "TO DO" comments look like good ideas to me.

And, the short Kassiopeia example runs correctly.

PR #88, opened 2019-09-23
Feature/code unification

pslocum:
New class by PTS FIRHandlerCore handles receive and transmit FIR operations, and is implemented in  signal generators.  LMCPowerCombiner is now more general and is also implemented.  Class variables have been moved out of signal generators and into suitable classes.  Indenting is improved.  Extra print statements in LMCDigitizer are removed.  Local test merge with develop has gone smoothly.


psurukuc:
Penny, one thing that didn't propogate over from the AntennaSignalGenerator was the bugfix I made for the AOIFactor in the commit e277548bb2c5c27e13dd310516ed885b95d8c6f7. We will need to make sure that it propogates into develop.

buzinsky:
looks a lot better, thanks

pslocum:
> Penny, one thing that didn't propogate over from the AntennaSignalGenerator was the bugfix I made for the AOIFactor in the commit [e277548](https://github.com/project8/locust_mc/commit/e277548bb2c5c27e13dd310516ed885b95d8c6f7). We will need to make sure that it propogates into develop.

Is this the code with the bug fix below?  I am finding it in the present pull request as-is?

`    double DipoleSignalGenerator::GetAOIFactor(LMCThreeVector TrasmitterReceiverNormal,LMCThreeVector ReceivingPatchNormal)
    {
	if(TrasmitterReceiverNormal==LMCThreeVector::sZero)
	{
	    return 1.0; //Unusual case where source is same as the receiver
	}
	double AOIFactor = fabs(TrasmitterReceiverNormal.Unit().Dot(ReceivingPatchNormal.Unit()));
        return AOIFactor;
    }`

PR #89, opened 2019-09-25
Feature/lienard wiechert

buzinsky:
Moved free field calculations out of duplicated code into separate class, which is called in patch generator, free field generator

Tested on FreeField generator only, does skip, find zeros, produce reasonable values. Tests not much more detailed than that


pslocum:
This is looking great!  I just merged develop into feature/LienardWiechert locally, resolved conflicts in LMCPatchSignalGenerator, and compiled it.  Do you mind if I push that merge to remote feature/LienardWiechert for more testing on the cluster?  And then after testing we merge the feature into develop.

buzinsky:
Thanks, that would be great

PR #95, opened 2019-10-08
write pitch angle in FTG to ROOT

buzinsky:
Can now add pitch angles (in deg.) to output of FTG

image


buzinsky:
(obviously I tested this to get the above output plot)

PR #97, opened 2019-10-08
Write random seed to Root file in FTG.  Improved parameter exception …

pslocum:
…message in LMCGaussianNoiseGenerator.

We can release a patch if these changes look reasonable and if they are needed soon.


buzinsky:
I also approve this: code looks good and it works when I test it. Thanks

PR #98, opened 2019-10-10
Added boolean flag to LMCFakeTrackSignalGenerator to switch pitch ang…

pslocum:
…le corrections on off.  In LMCGaussianNoiseGenerator::Configure(), return false if noise power not configured.


buzinsky:
this looks all good. Could you set the initialization and the conditional line from "1" to "true", just to be a little clearer its of different type?

PR #99, opened 2019-10-18
Feature/unify lo

buzinsky:
Remove LO from being included in generators: separated as own generator

Tests:
This seemed to work when I tried it on the FTG: Attached is a picture of tracks at the correct frequency when this is used (though I suppose there is some possibility it was aliased _just_ to the correct frequency)
Screen Shot 2019-10-08 at 4 42 39 PM

### DIFF
--- a/Source/Generators/LMCDipoleSignalGenerator.cc
+++ b/Source/Generators/LMCDipoleSignalGenerator.cc
@@ -23,7 +23,6 @@ namespace locust
     DipoleSignalGenerator::DipoleSignalGenerator( const std::string& aName ) :
     Generator( aName ),
     fDoGenerateFunc( &DipoleSignalGenerator::DoGenerateTime ),
-    fLO_frequency( 20.05e9 ),
     fRF_frequency( 20.1e9 ),
     fArrayRadius( 0. ),
     fNPatchesPerStrip( 1 ),
@@ -34,7 +33,6 @@ namespace locust
     EPhaseBuffer( 1 ),
     EAmplitudeBuffer( 1 ),
     EFrequencyBuffer( 1 ),
-    LOPhaseBuffer( 1 ),
     IndexBuffer( 1 ),
     PatchFIRBuffer( 1 ),
     fFieldBufferSize( 50 )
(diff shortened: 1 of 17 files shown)


buzinsky:
I'm not quite sure I'm following these questions. The idea of the generator, is that it applies the LO after the signals have already been generated: acts on the signal array.
I do like it as is as its own generator: its not clear to me why we'd want a second class.

3. Im fine with the idea: Im not sure it would work, right? If LMCFTG calls GetLOFrequency() before LO Class is instantiated, can it even call that function?

pslocum:
The problem for right now is that the FTC can't write the LO frequency to the output Root file because it doesn't know what it is.  If we had an LO class in RxComponents we could Configure() it in the FTC, get the LO frequency, and then do the downmixing later.  Or we could find some other way to write the LO frequency to file in the FTC.

buzinsky:
ok, ya. The FTG is a special case here, because we want to write it to a ROOT file. I dont think there are any other generators, which are expected to have this problem going forward, right?

I can think more about what is the least invasive way to resolve this

pslocum:
yes, right, it is just the FTG at the moment.

buzinsky:
I think if this is really the only generator that we expect this to happen in near future, (eg writing to ROOT file), I am okay with having a LO parameter in the FTG that solely writes to file, while still having one use the LO generator to do the actual downmixing

pslocum:
It seems like a feature that could cause a subtle problem on occasion in that there could be 2 different LOs.  It would probably be fine most of the time, but ...

buzinsky:
Right that is the concern: if the user makes a mistake, and causes the two LO config values to be misaligned, that would cause incorrect results. 

I think if we introduce an LO generator, but decide here that "ok lets have the FTG have an internal LO, so it can also write it to a ROOT file", you are just as likely to get a user error where they apply two LOs, which will also be wrong. 

Maybe the solution to this exception is to instead think harder about the ROOT file interface, not the LO itself

buzinsky:
its an unfortunate case, as I think it does somewhat limit the appeal of unifying/ removing all these LOs from the arguments of the various generators

pslocum:
If we had a mixer generator (instead of an LO generator), then we could have a separate LO component that is not a generator.  The LO frequency could then be defined once.  ?

pslocum:
The recommendations in this pull request can likely be integrated into the planned nymph upgrade to Locust.  Therefore we can close this for right now and then revisit.

PR #100, opened 2019-10-18
Feature/min pitch

buzinsky:
Alterations to FTG. Goal: allow one to end events based on pitch angle (as well as specification).

Setting min-pitch enforces the minimum pitch angle, at which events are cut short. Event length is totally set by this in that case. 
Setting ntracks-mean to non-zero overrides this, number of tracks is totally set by geometric distribution.

I made some changes to DoGenerate(). There was an error where the multi channel sim would not work as fStartFreq was never reset. Now I have fStartFreq only changes on creation of new tracks, not due to radiation losses. 

In both cases, (ntracks = geometric, or from pitch angle), I get sensible tracks, but a good check is appreciated.

### DIFF
--- a/Source/Core/LMCEvent.cc
+++ b/Source/Core/LMCEvent.cc
@@ -19,6 +19,21 @@ namespace locust
 {
     Event::Event() {}
     Event::~Event() {}
+
+    void Event::AddTrack(const Track aTrack)
+    {
+        StartFrequencies.push_back( aTrack.StartFrequency );
+        TrackPower.push_back( aTrack.TrackPower );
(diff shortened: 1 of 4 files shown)


pslocum:
I can see the need to advance the fCurrentFrequency and not the
fStartFrequency.  Does this line:
tTimeOffset = fEndTime; mean that there is no delay between tracks, or is
there some other explanation?

buzinsky:
Yes, that means between tracks, there is no delay. 
Was there supposed to be the 0.5 ms there? I realize I may have misinterpreted it as Im typing this

PR #102, opened 2019-10-21
remove confusing line (w/ no effect)

buzinsky:
Remove the 0.5 ms spacing (which had no effect anyways)

Tracks now have no spacing, events are distributed randomly


buzinsky:
> So the line
> `tTimeOffset += tEventSpacing;`
> is removed without effect because the tTrackIndex loop breaks immediately afterward, and a new Event is then started with a randomized start time.

Yes, exactly

> 
> So we now have randomized start times for events with no time between tracks, by way of
> `tTimeOffset = fEndTime; SetTrackProperties(aTrack, tTrackIndex, tTimeOffset);`
> If that sounds correct, should we do the release now?

That sounds correct. I am okay with releasing this

PR #111, opened 2019-11-14
Feature/inverse fft implementation

psurukuc:
Implementation of TFHandler which automatically converts Transfer functions to Finite Impulse Response within Locust.


pslocum:
These changes are tested and will improve our handling of HFSS output in Locust.

PR #119, opened 2020-01-07
Feature/ftg coupling fix

buzinsky:
@cclaessens pointed out that the coupling of the waveguide in the FTG falls off very slowly, much more than observed in reality (where ~<89 deg. is unobserved)
power_vs_angle

The error was that the fTrapLength (corresponding to L0 in the phenomenology paper) was set to the coil radius, as opposed to the parabolic coefficient describing the field (though it is proportional to the coil radius, for a harmonic trap).

I worked out that they are related via: `L_0 = R  sqrt(2/3) sqrt((B_bkg-B_coil) / B_coil) `, where `B_coil = mu_ N I / 2R`, the magnetic field at the center of a current loop. For Phase II, R = 0.0084 m (https://github.com/project8/locust_mc/blob/master/Config/Tutorial/Project8Phase2Geometry.xml) and N = 63 turns (https://github.com/project8/locust_mc/blob/master/Config/Tutorial/Project8Phase2_WithRoot_Template.xml). This gives me the 0.1784 m value. This can be also seen (to about 10%) by directly eyeballing the following picture and  estimating L0, given B(z) = B0 (1 + (z/L)^2 ) (by defn).

image

Finally, at 0.1784m, the bessel function coupling will look like this: 
New Coupling

I opted to leave L0 directly configurable, as opposed to the (many) parameters that comprise it, even if they are more physical


buzinsky:
I think the option is to either specify L0, as we have, or replace that with 3 additional parameters: coil radius, coil current, wrapping number of the coil, (we already have background field), from which L0 is automatically calculated. Now that I think about it, the 2nd path is much easier to use, even though it means even more parameters

PR #123, opened 2020-01-17
Implemented signal summing for fPowerCombiner = slotted-waveguide in …

pslocum:
…LMCPowerCombiner::SetVoltageDampingFactors().  Generalized npatches-per-strip and patch-spacing to nelements-per-strip and element-spacing in LMCPlaneWaveSignalGenerator, LMCPatchSignalGenerator, LMCTurnstileSignalGenerator, LMCDipoleSignalGenerator, and LMCFreeFieldSignalGenerator.

Left to-do:    Define single slot pattern in LMCSlotAntenna.


psurukuc:
Thanks for doing this Penny. So, the plan is to keep the name of the classes unchanged for now ?

pslocum:
Right now we have the slotted waveguide selectable by parameter in
LMCPowerCombiner, with individual slot characteristics defined with
LMCReceiver.  So these changes are pretty minimal.  I do have some thoughts
on classes to unify on the transmit side, but they would be in a different
pull request.  Maybe we can chat on Tuesday.

pslocum:
With these changes the LMCArraySignalGenerator is switching from patches to slots with the "power-combining-feed" value in the config file.  If possible, I would like to merge this into develop to keep up with ongoing development on the slotted waveguide antenna in another repo branch.

PR #130, opened 2020-02-11
Moved Kass field solutions out of LMCArraySignalGenerator to LMCKassT…

pslocum:
…ransmitter.  Started to move Kass-dependent classes out of non-Kass build in Source/CMakeLists.txt.

This update continues toward removing LMCTransmitter functionality from LMCArraySignalGenerator.  We still have to build with Kassiopeia even if the LMCTransmitter is not Kassiopeia, but the LMCKassTransmitter calculations are now done in a subclass of LMCTransmitter.  This makes the LMCArraySignalGenerator easier to read and maintain.  This modification runs on my laptop but is still being tested on the cluster.


Source/CMakeLists.txt
```diff
@@ -113,6 +112,7 @@ if (locust_mc_BUILD_WITH_KASSIOPEIA)
         Generators/LMCArraySignalGenerator.hh
         Generators/LMCKassSignalGenerator.hh
 
+        Transmitters/LMCKassTransmitter.hh
```
psurukuc: Should line 95 be removed as well ?

psurukuc:
Penny, I have a comment. When I created the transmitter, in my mind I had this notion that the transmitter is independent of the receiver. As I look at the code closely, I notice that the transmitting antenna is now dependent on the receiver via the currentElement in GetEFieldCoPol function. This circular dependency where the receiver needs an antenna and the antenna uses a receiver might become a problem in long term. I could think of scenarios where for example we want to just plot the fields (maybe sometime down the line) from a turnstile antenna at several locations in the detector. The way it is defined currently forces us to define an array. If possible I propose the code to be such that the receiver knows about the transmitter's existence but the transmitter's only job is to transmit from a location and it doesn't know about what is receiving the signal as in real life. And the functionality within the transmitter that needs the receiver is not too much and could be moved into the ArraySignalGenerator.

pslocum:
Right, I can see there is some possible overlap between the transmitter and
the array, in that we are selecting the copolar/crosspolar components of
the transmitted fields at each array element.   I agree we could be doing
that selection in the LMCArraySignalGenerator instead of in the
LMCTransmitters.  It should be straightforward to fix.

pslocum:
Merging this into develop for now, but keeping branch to finish unifying LMCTurnstileSignalGenerator.

PR #132, opened 2020-02-19
New parent class LMCTransmitterHardware to describe orientation and a…

pslocum:
…ngular dependence of transmitting antennas, e.g. LMCDipoleAntenna, used in LMCAntennaSignalTransmitter.  Deleted LMCTurnstileSignalGenerator.

These changes are intended to unify the LMCTurnstilSignalGenerator with the other LMCTransmitters.
It compiles but does not yet run.  I am opening the pull request for discussion in the meantime.


Source/Transmitters/LMCAntennaSignalTransmitter.cc
```diff
@@ -77,7 +77,12 @@ namespace locust
     
     bool AntennaSignalTransmitter::SetAntennaType( std::string transmitterAntennaType )
      {
-     	if (transmitterAntennaType == "antenna-signal-dipole") fAntennaType = 0; // default
+     	if (transmitterAntennaType == "antenna-signal-dipole")
+     		{
+     		fAntennaType = 0; // default
+    		DipoleAntenna* modelTransmitterHardware = new DipoleAntenna;
```
psurukuc: Couldn't these two lines be combined into a single line like this ?
 fTransmitterHardware  = new DipoleAntenna;
since TransmitterHardware is a parent class of  DipoleAntenna

Source/Transmitters/LMCDipoleAntenna.cc
```diff
@@ -11,7 +11,8 @@
 namespace locust
 {
 
-    DipoleAntenna::DipoleAntenna()
+    DipoleAntenna::DipoleAntenna():
+    fMomentVector( 0., 0., 1.0 )
```
psurukuc: I think it makes sense for this to be configurable. For example, this moment is (1,0,0) in Maurio's dipole case.

Source/Transmitters/LMCDipoleAntenna.cc
```diff
@@ -20,4 +21,18 @@ namespace locust
     }
 
 
+    void DipoleAntenna::TxHardwareSayHello()
+     {
+    	printf("Dipole says hello\n"); getchar();
+     }
+
+    double DipoleAntenna::GetPatternFactor(LMCThreeVector pointOfInterest)
```
psurukuc: Thanks for implementing this.

pslocum:
This is probably ready for the default antenna orientations and angular dependence to be corrected in LMCDipoleAntenna and LMCTurnstileAntenna.  I am not sure what the best values are but have left placeholder vectors and angular dependence as a starting point.

psurukuc:
@pslocum, as it turns out since you opened the PR, I can't assign you as a reviewer. Please take a look at the changes and see if you find anything odd. Please pay particular attention to the PlanewaveTransmitter since I put some extra logic there that I think works but not a 100% sure.

pslocum:
I see, so you are doing InitializeFieldPoints(allRxChannels) for all of the
LMCTransmitters, which means they don't need a pointOfInterest at every
sample.  That looks good.  I also tested the LMCPlaneWaveTransmitter, and
its gain and symmetry look correct to me.  I am ready to merge this into
develop if you are.  Thanks!

PR #135, opened 2020-02-25
Feature/subarrays

pslocum:
Minor change to allow for longitidinal subarrays each with its own digitizer.

### DIFF
--- a/Source/Generators/LMCArraySignalGenerator.cc
+++ b/Source/Generators/LMCArraySignalGenerator.cc
@@ -412,7 +412,7 @@ namespace locust
         const double elementRadius = fArrayRadius;
         double zPosition;
         double theta;
-        const double dThetaArray = 2. * LMCConst::Pi() / nChannels; //Divide the circle into nChannels
+        const double dThetaArray = 2. * LMCConst::Pi() / (nChannels/nSubarrays); //Divide the circle into nChannels
         const double dRotateVoltages = 0.;  // set to zero to not rotate element polarities.
 
         allRxChannels.resize(nChannels);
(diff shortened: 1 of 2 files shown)


Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -412,7 +412,7 @@ namespace locust
         const double elementRadius = fArrayRadius;
         double zPosition;
         double theta;
-        const double dThetaArray = 2. * LMCConst::Pi() / nChannels; //Divide the circle into nChannels
+        const double dThetaArray = 2. * LMCConst::Pi() / (nChannels/nSubarrays); //Divide the circle into nChannels
```
psurukuc: Is this right ? From the code it looks like the total number recievers(patches/slots) = nsubarray*nchannels*nelementsPerchannel. But, this line seems to divide the circumference of array by number of channels and multiply by the number of arrays.

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -409,28 +417,33 @@ namespace locust
 
         allRxChannels.resize(nChannels);
 
-        for(int channelIndex = 0; channelIndex < nChannels; ++channelIndex)
+        for (int subarrayIndex = 0; subarrayIndex < nSubarrays; ++subarrayIndex)
```
psurukuc: To me it makes sense to do:
for each channel:
  for each subarray:
    for each receiver: 
That way, the theta can still be calculated accurately. 
On a side note, we should also think about how this affects the reconstruction with Katydid.

pslocum:
I think that your idea might be a better way to do it, but in the short
planning time the reconstruction seemed like it would need some
development.  By doing it the above way I am trying to do a coarse quick
test:  3 separate independent subarrays with nChannels/nSubarrays each.  So
in the reconstruction we would just focus each subarray (one centered at
some -z with nChannels/nSubarrays, one centered at zero, and one centered
at some +z, extract the signal power from each subarray like usual, and
then sum the powers.  The reconstruction will need to be checked outside of
Katydid for now.  In the config file I am picking n-subarrays = 3,
n-channels = 90, nelements-per-strip=10.  Channels 0-29 are the 1st
subarray at some -z, 30-59 the second centered at 0, and 60-89 are the
third at some +z.

It is still running on the Grace cluster, so it might still be wrong...

pslocum:
I am going to cross check this approach with summed parallel simulations
each with a shifted small array.  It should be the same.

pslocum:
Thinking to merge this today.  It is not a final subarray configuration but it does produce one egg file with results identical to the multi-egg file workflow.  And the default n-subarrays: 1 is still equivalent to not having any subarrays.

PR #138, opened 2020-03-23
Feature/separateroot

pslocum:
…RootTreeWriter is a singleton class for access by multiple LMCGenerators.

This feature branch is underway and is not yet ready for merge.  Its purpose is to separate the Root writer(s) from the LMCGenerators without requiring any immediate modifications to config files.  The remaining step in this feature will be to write a singleton LMCRootHistogramWriter.  Question:  Should the base class LMCFileWriter also be a singleton, or is it acceptable for only the subclasses to be singletons?  It works correctly as implemented in that all LMCGenerators can access the same static LMCRootTreeWriter instance.

This modification will help to allow for separating the downmixing functionality into its own LMCGenerator as in https://github.com/project8/locust_mc/pull/99 .


psurukuc:
> …RootTreeWriter is a singleton class for access by multiple LMCGenerators.
> 
> This feature branch is underway and is not yet ready for merge. Its purpose is to separate the Root writer(s) from the LMCGenerators without requiring any immediate modifications to config files. The remaining step in this feature will be to write a singleton LMCRootHistogramWriter. Question: Should the base class LMCFileWriter also be a singleton, or is it acceptable for only the subclasses to be singletons? It works correctly as implemented in that all LMCGenerators can access the same static LMCRootTreeWriter instance.
> 
> This modification will help to allow for separating the downmixing functionality into its own LMCGenerator as in #99 .

I think a subclass can be singleton without the parent class being one.

buzinsky:
You certainly could do that (have non-singleton base & singleton derived). I think my question to you is what else could we conceivably want to inherit from file writer besides this? And would these new writers have to also be singletons as a matter of necessity?

Because, if they do, then instead of copying over code each time we create a new derived class, my impression is that making the base singleton will save work, so that all the derived are also singletons (http://www.yolinux.com/TUTORIALS/C++Singleton.html). I have never tried this, so not sure how it would work out

Basically, do any of the classes, (or all) of the writers need to be singletons? I thought the thing we wanted was for the writers to have object permanence between generators, and not necessarily uniqueness. The exact relationship between base & inherited classes may be easier once we create/ initialize some object w/ the right lifetime

nsoblath:
There's a base class in scarab for making classes singleton: `library/utility/singleton.hh`

Steps for use:
        1. Inherit your class from `singleton< your_class >`
        2. Make your constructor and destructor protected (or private)
        3. Add the `allow_singleton_access( your_class )` macro to your class definition to allow the base classes to access your class

pslocum:
Thank you everyone for the ideas.  Base class LMCFileWriter is now a
scarab::singleton .  I will add some more writer subclasses, run some more
tests, and then request another review.

PR #139, opened 2020-03-25
Feature/config distributions

buzinsky:
Not quite ready to merge, though can be looked at & tested

This branch allows you to change randomly generated distributions at runtime, eg so you can change your config file to change the slope to be drawn from uniform, gaussian, etc distributions

The only outstanding things (that I am aware of) right now are:
a) I need to swap out the barycentric interpolator for something else, I am getting oscillations in the interpolator (analogous to https://en.wikipedia.org/wiki/Runge%27s_phenomenon), nothing in boost is sufficient. Likely swapping to GSL
b) A parameter for the seed should be passed to all generators
c) Code messiness / formatting (mostly in KrComplexLine)

I am worried that each derived class having its own generator (probably with the same seed) means that random numbers between different instances will be completely correlated, which could subtly bias results. Attached is a FTG sample config that seems to work (github doesnt like json extension)
If we want to add more distributions or let more variables be distributions in the FTG or elsewhere, please let me know

[LocustFakeTrack2.txt](https://github.com/project8/locust_mc/files/4382455/LocustFakeTrack2.txt)


pslocum:
This looks like it will be useful.  
I did get a compile error on Ubuntu.  Is there a Boost upgrade that is needed?

`/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:`` In member function ‘std::valarray<double> locust::KrComplexLineDistribution::P_1s_nprime(const double&, const std::valarray<double>&)’:
/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:149:37: error: no matching function for call to ‘pow(std::valarray<double>&, int)’
         P_1s_nprime *= pow(n_prime,8);`

buzinsky:
We shouldn't need a boost upgrade, I couldnt see any boost interpolators which were sufficient for these purposes. I finished the change to GSL today, which works very nicely. I will need your help changing whatever CMake is needed to link it. I just used the command: 
`cmake ../locust_mc/ -DCMAKE_CXX_STANDARD_LIBRARIES=“-lgsl”` to get it working

Thanks for the comment: I think your compiler is pickier than mine for the type conversions here, as that wasn't an issue mine. I can try to fix that

Since I made the original comment resolved a) b), and the concern about too many RNG engines, still need to point c)

buzinsky:
I think that should fix it, in the documentation it doesnt like if the sizes of pow(base, exp) are the different, so I need to explicitly do pow(array1, array2), not pow(array, number). 
When you try again, I think it should work, or if I missed another one of these, you may need to do something similar. The valarrays are nicer than the vectors, in that you can apply algebra directly to them.

I dont mean to keep you up, I am out west. I think I can finish formatting tonight

psurukuc:
The code in general looks good to me. I tested and I get a new but related error with valarray. I think Penny and I use the same compiler. I attach the error if it helps you debug. Let me know if you need help with debugging since you don't see the error on your end. 
[debug.txt](https://github.com/project8/locust_mc/files/4388408/debug.txt)

psurukuc:
I will also add this valarray seems really helpful. @pslocum I think this might be a way for us to speed up the convolution.

buzinsky:
Ya, the only downside is that you cant change the size dynamically like vectors.

I think if you or penny could help with a bit with these compile errors, that would be fantastic
I think the two things to try would be go to (8. * ones ) (so its double x valarray<double> instead of int)
if that doesnt work: (ones * 8.). sometimes the overloaded operators work better this way. Not totally sure

buzinsky:
> I will also add this valarray seems really helpful. @pslocum I think this might be a way for us to speed up the convolution.

My impression was that this was only really useful for making the code succinct, instead of doing, for example, a loop to multiply all the elements by 8, if your compiler lets you, the code is much shorter. It still is going to internally do a loop, which may be slightly optimized, but both are going to be O(N) time cost

pslocum:
In the above I had not linked to GSL in the compiling, so that may have been a problem.  But now when I cmake like this:
cmake ../ -DCMAKE_CXX_STANDARD_LIBRARIES=-lgsl

the compile stops with this:
//usr/local/lib/libgsl.so: undefined reference to `cblas_ztrsv'

The gsl version I have appears to be 2.5 .

pslocum:
And cmake without -lgsl reproduces the same problem, so it looks like the
gsl linking is not happening on my laptop at the moment.

buzinsky:
Even with the quotation marks like I had? Or if you know where to add the -lgsl in the cmake files themselves, that is the long term solution

I am running gsl 2.6  . I guess it could be, but I do kind of doubt that that small of a difference is the issue here: I am just doing basic interpolation, I think it shouldn't be something brand new for them

the gsl is unrelated to valarray issues

pslocum:
with the quotation marks on command line:
[ 24%] Linking CXX shared library libyaml-cpp.so
c++: error: “-lgsl”: No such file or directory

without the quotation marks on command line:
[ 53%] Linking CXX executable M3Info
//usr/local/lib/libgsl.so: undefined reference to `cblas_ztrsv'

setting CMAKE_CXX_FLAGS in ccmake to -lgsl:
[ 53%] Linking CXX executable M3Info
//usr/local/lib/libgsl.so: undefined reference to `cblas_ztrsv'

setting CMAKE_CXX_FLAGS in ccmake to -lfaketest:
[ 24%] Linking CXX shared library libyaml-cpp.so
/usr/bin/ld: cannot find -lfaketest
collect2: error: ld returned 1 exit status

pslocum:
I got past the gsl problem with
cmake ../ -DCMAKE_CXX_STANDARD_LIBRARIES=-lgsl
-DCMAKE_CXX_STANDARD_LIBRARIES=-lgslcblas -DCMAKE_CXX_STANDARD_LIBRARIES=-lm

and then crashed with this:
[ 69%] Building CXX object
Source/CMakeFiles/LocustMC.dir/Distributions/LMCKrComplexLineDistribution.cc.o
/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:
In member function ‘std::valarray<double>
locust::KrComplexLineDistribution::P_1s_nprime(const double&, const
std::valarray<double>&)’:
/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:137:37:
error: no match for ‘operator*’ (operand types are ‘int’ and
‘std::valarray<double>’)
         P_1s_nprime *= pow(n_prime,8*ones);
                                    ~^~~~~

buzinsky:
Great, I think this might help, if you could try it:
```I think if you or penny could help with a bit with these compile errors, that would be fantastic
I think the two things to try would be go to (8. * ones ) (so its double * valarray instead of int)
if that doesnt work: (ones * 8.). sometimes the overloaded operators work better this way. Not totally sure```
with the 8. changing the 8 from an int, to a double, I hope

pslocum:
P_1s_nprime *= pow(n_prime,8.*ones); crashes with:
/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:138:43:
error: no matching function for call to ‘pow(std::valarray<double>&,
std::_Expr<std::_BinClos<std::__multiplies, std::_Constant, std::_ValArray,
double, double>, double>)’
         P_1s_nprime *= pow(n_prime,8.*ones);
                                           ^
P_1s_nprime *= pow(n_prime,ones*8); crashes with:
/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:138:40:
error: no match for ‘operator*’ (operand types are ‘std::valarray<double>’
and ‘int’)
         P_1s_nprime *= pow(n_prime,ones*8);

P_1s_nprime *= pow(n_prime,ones*8.); crashes with:
/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:138:43:
error: no matching function for call to ‘pow(std::valarray<double>&,
std::_Expr<std::_BinClos<std::__multiplies, std::_ValArray, std::_Constant,
double, double>, double>)’
         P_1s_nprime *= pow(n_prime,ones*8.);

buzinsky:
thanks for trying, this is stumping me… the other things I can think of would be to just check if you put in the line std::valarray<double> c = 8 * ones; right after the ones is defined if, that works or not. If that line works, it means the multiplication is not the problem

Otherwise, going to `std::pow` instead of just `pow` is probably good practice anyways. Tough to debug w.o access

pslocum:
It looks like the multiplication is the problem:

/home/penny/locust_mc/Source/Distributions/LMCKrComplexLineDistribution.cc:136:37:
error: no match for ‘operator*’ (operand types are ‘int’ and
‘std::valarray<double>’)
         std::valarray<double> c = 8 * ones;
                                   ~~^~~~~~

psurukuc:
Works for me. Here is what I did:
        P_1s_nprime *= pow(n_prime,std::valarray<double>(8.0*ones));
        P_1s_nprime *= pow( pow(n_prime, std::valarray<double>(2.0*ones)) + 1., std::valarray<double>(-4.0*ones));
        P_1s_nprime *= exp(-4. * n_prime * atan(1. / n_prime));

buzinsky:
[LocustFakeTrack.txt](https://github.com/project8/locust_mc/files/4390456/LocustFakeTrack.txt)

Thanks for working through the compilation issues: I should mention that if a certain distribution is not assigned a distribution (or the name is misspelt), the object doesnt get assigned, and there is a segfault

I need to put in default distributions & logger messages to make this more user accessible. attached is the config, that I have working. you can swap distribution out if you like

buzinsky:
Some plots from printing variables from the KrComplexLineDist (which is the biggest/ most susceptible to a bug):
total kr line shape:
KrTotalShape

shake spectrum monte carlo 
shake_sp_MC

compare with Yu Haos presentation https://3.basecamp.com/3700981/buckets/3107037/uploads/2439975441 slide 3

Krypton energy loss CDF (eg: randomly generate uniform number u [0,1], graph tells you energy loss). Note interpolation handles excitation region (resonances) at ~10-15 eV. Agrees with PDF
krCDF

buzinsky:
ya, when I try a fresh build without any modifications (eg. simple `cmake ../locust_mc`). I get a linker error at 97%, even after pulling. This doesnt happen on linux if you do it? (kassiopeia is off)

pslocum:
right, if I do a make clean and git pull, then ccmake ../ , configure,
generate, make install, it compiles and runs without Kassiopeia with
generates normal-looking FTG output.

I have these lines in my ccmake window:
pkgcfg_lib_GSL_gsl               /usr/local/lib/libgsl.so
 pkgcfg_lib_GSL_gslcblas          /usr/local/lib/libgslcblas.so
 pkgcfg_lib_GSL_m                 /usr/lib/x86_64-linux-gnu/libm.so

buzinsky:
Hmm, I dont see those lines anywhere.
Sorry, can I ask you to try again? The reason is that if I do a cmake with the gsl library (using these extra command line commands),  then `make install`, if I do  `make clean`, `cmake ../locust_mc` , and `make install` it works, even if just a pure `cmake ../locust; make install` doesnt. To my knowledge `make clean` doesnt actually remove the cmake information. If you could start with a empty build directory and try to compile that would be great.

I think if there is an issue it should come up at the end of compilation & not during running. Thanks again

pslocum:
Good idea.  So I deleted the cbuild directory with rm -rf cbuild/ and then
recreated it, reconfigured, and compiled and ran successfully.  But you are
right, the cached cmake information was changed by this and no longer
mentions anything about gsl.

You have just the one -lgsl flag, correct?

buzinsky:
Very interesting: ya to get it to work, I need to just add that flag.
`cmake ../locust_mc/ -DCMAKE_CXX_STANDARD_LIBRARIES="-lgsl"`

buzinsky:
Hey, can you try one more time from an empty directory?
I may not have been so helpless here: I was able to guess & figure out the changes to the cmake just by using context. It works for me now (and I really hope it works for you as well)

This would make gsl a compilation requirement. I can try to look up the min. version for the interpolators, and update the README. If this looks good to you I'm going to run last minute tests tonight, add some loggers. I'll send a message to give an all clear on the merge

psurukuc:
It works for me with a fresh build into an empty directory. I tried after your recent changes in CMakeLists.

Pranava Teja Surukuchi
__________________________________
Postdoctoral Research Associate
Wright laboratory, Yale university
Email: pranavateja.surukuchi@yale.edu

pslocum:
Yes, making GSL required works for me too after git pull and rm -rf cbuild/
.  And, I can see the gsl information in the ccmake window, and it is now
printed to CMakeCache.txt .

buzinsky:
I'm giving you the green light to merge: I'm not seeing anything else that's clearly wrong either visually in the code, or in my tests. 
Thanks again to both of you for reviewing this quickly in the last few days

PR #140, opened 2020-03-27
Hotfix/reduce convolution speed

psurukuc:
Not ready for merge yet but will work on this a bit more to try to smartly allocate the size of FIR filter. Currently, a modification of looping through the bins to increase the speed of convolution. Should increase the speed by a few % (nothing jaw dropping).


psurukuc:
@arinabykadorova, when you get a chance could you check if this increases the speed a bit. Nothing urgent though.

pslocum:
This runs as expected and looks good.  Thank you for finding this improvement.

PR #146, opened 2020-05-28
Added debug statements to print out voltages in order to check digiti…

pslocum:
…zer range.

This will allow for more routine checking of the digitizer range when cmake_build_type is set to "debug".  For faster running the cmake_build_type will need to be set to "release".

### DIFF
--- a/Source/RxComponents/LMCPowerCombiner.cc
+++ b/Source/RxComponents/LMCPowerCombiner.cc
@@ -78,7 +78,7 @@ namespace locust
 		aSignal->LongSignalTimeComplex()[sampleIndex][1] += 2.*VoltageFIRSample * cos(phi_LO);
 
 		if ( (sampleIndex%100 < 1) && (fvoltageCheck==true) )
-			LDEBUG( lmclog, "Voltage " << sampleIndex << " is <" << aSignal->LongSignalTimeComplex()[sampleIndex][1] << ">" );
+			LWARN( lmclog, "Voltage " << sampleIndex << " is <" << aSignal->LongSignalTimeComplex()[sampleIndex][1] << ">" );
 
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


Source/RxComponents/LMCPowerCombiner.cc
```diff
@@ -71,8 +77,8 @@ namespace locust
 		aSignal->LongSignalTimeComplex()[sampleIndex][0] += 2.*VoltageFIRSample * sin(phi_LO);
 		aSignal->LongSignalTimeComplex()[sampleIndex][1] += 2.*VoltageFIRSample * cos(phi_LO);
 
-		if (sampleIndex%100 < 1)
-    		LDEBUG( lmclog, "Voltage " << sampleIndex << " is <" << aSignal->LongSignalTimeComplex()[sampleIndex][1] << ">" );
+		if ( (sampleIndex%100 < 1) && (fvoltageCheck==true) )
```
psurukuc, line 80: Looks, good. This might be an issue if the frequency is a (or close to) multiple of 100.

arinabykadorova: Thanks for doing this, Penny. Will It be an issue that this occurs before the low-pass filter? I usually print voltages from LMCDecimateSignalGenerator because that's the last thing that happens to them before they go to the digitizer. If that isn't an issue, then this is fine.

pslocum:
Am thinking to leave the voltage-check option in LMCPowerCombiner so that
we don't have to wait until the end of the simulation to see how they
look.  Otherwise it could be a long wait for the
LMCDecimateSignalGenerator.  The possible 3 dB reduction after the LPF can
be explained in the documentation.  The digitized voltages printed to
terminal are a more final look at the voltages if noise is included.  We
can also say in the documentation that the most stringent way to check the
digitizer range is to plot a time series after the simulation.

PR #149, opened 2020-06-17
Removed duplicate declaration of gas_species in LMCKrComplexLineDistr…

pslocum:
…ibution.cc

(Is there a better way to handle this such that gcc does not interpret this as a duplicate declaration?)


pslocum:
I think it got past me during a merge conflict resolution.

PR #151, opened 2020-07-16
Feature/ftg 20 update

buzinsky:
I think this is working now. Upgrades to FTG for SA group:

-Converted track length distributions to the run-time distributions (with assumption most people will keep exponential)
-More realistic start pitch angle distribution. You generate initial position of decay in z0 parameter (likely uniform), then it converts to pitch at bottom
-I also made the trap length configurable


pslocum:
Does this update need to have a new release?

buzinsky:
Yes, please: these were all the changes the group wanted

PR #154, opened 2020-07-29
Updates to Slotted Waveguide to be compatible with new 5-slot TF

arinabykadorova:
Removed the scaling factor that was used for the 10-slot array. Changed the default impedance transformation for 390 ohms, the input impedance of the array for the 5-slot transfer function. (square root of 50 over 390). Note that if you want to use the 10-slot TF, you can only use it with 10 slots, and you would have to specify the impedance transformation in the config file.


pslocum:
Is there a new 5-slot transfer function file that should accompany these changes, and is the plan to push it to the hexbug repo?

arinabykadorova:
Yes, there is a new TF. It's on hexbug on develop. I have tested it with this code.

PR #164, opened 2020-09-14
Added AM modulation/demodulation option to LMCTestSignalGenerator.

pslocum:
Expanded some config options in LMCTestSignalGenerator for initial demodulation tests.

### DIFF
--- a/Source/Generators/LMCTestSignalGenerator.cc
+++ b/Source/Generators/LMCTestSignalGenerator.cc
@@ -25,6 +25,8 @@ namespace locust
         fDoGenerateFunc( &TestSignalGenerator::DoGenerateTime ),
         fLO_frequency( 20.05e9 ),
         fRF_frequency( 20.1e9 ),
+		fAMModfrequency( 0.0e6 ),
+		fAMModSelect( 0 ),
         fAmplitude( 5.e-8 ),
 		fMixingProduct( false ),
     	fWriteRootHisto( false ),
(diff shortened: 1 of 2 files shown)


Source/Generators/LMCTestSignalGenerator.cc
```diff
@@ -50,7 +52,14 @@ namespace locust
...
             SetLOFrequency( aParam.get_value< double >( "lo-frequency", fLO_frequency ) );
         }
-
+        if( aParam.has( "am-frequency" ) )
+        {
+            SetAMModFrequency( aParam.get_value< double >( "am-frequency", fAMModfrequency ) );
+        }
+        if( aParam.has( "am-mod-select" ) )
+        {
+            SetAMModSelect( aParam.get_value< int >( "am-mod-select", fAMModSelect ) );
```
psurukuc, line 61: Penny, what is fAMModSelect ? It is not clear to me.

pslocum: It is a 3-way flag, with values:
0 -> omit all AM modulation/demodulation (default value, backwards compatible with original LMCTestSignalGenerator.)
1 -> apply AM modulation to test signal with frequency "am-frequency"
2 -> apply AM modulation to test signal with frequency "am-frequency", and then demodulate to recover the unmodulated signal.

It gets applied in pow(cos(AMMod_phase),fAMModSelect) so that the test signal gets multiplied by either 1.0 for fAMModSelect = 0, a sinusoid cos(AMMod_phase) for fAMModSelect = 1, or cos^2(AMMod_phase) for fAMModSelect = 2.  This is following the steps in slides 8-13 in http://web.mit.edu/6.02/www/s2012/handouts/14.pdf .

Issue #116, opened 2019-12-20
Merge functionality of RawReverseFFT and ReverseFFT


psurukuc:
The current functionality of ReverseFFT in ComplexFFT could be generalized to include a generic forward and reverse FFT functionality that could be used for any kind of Fourier and reverse Fourier transforms to be used in Locust. Currently, the IFFT (i.e., reverse FFT ) only generates the real part and not the imaginary part. The other parts that could have improvements include generalizing the prebuffer size, the window size, and the zero-padding size.

pslocum:
This is addressed in https://github.com/project8/locust_mc/pull/165 .

PR #165, opened 2020-09-18
Instantiated ComplexFFT class in LMCHilbertTransform.  Compiles but n…

pslocum:
…o testing at all yet.

This is still being tested.  The purpose is to use only the ComplexFFT class for FFTs.  Will test fftw flags next.


psurukuc (approved):
All looks good to me @pslocum. Did you test and see if it gives identical results ?

pslocum:
> All looks good to me @pslocum. Did you test and see if it gives identical results ?

Right, the results look identical for e.g. the LocustPlaneWaveTemplate example, plots below:

and, run times are similar:
real	3m49.225s
user	3m48.663s
sys	0m0.509s

time for develop:
real	4m3.713s
user	4m3.251s
sys	0m0.392s

unifyFFT
develop

pslocum:
This addresses https://github.com/project8/locust_mc/issues/116

PR #166, opened 2020-10-01
Moved fftw_destroy_plan from ~ComplexFFT() to ForwardFFT() and Revers…

pslocum:
…eFFT().  This plugs a memory leak that has recently arisen due to new repetive use of ComplexFFT class.


arinabykadorova (approved):
Checked the pattern with this, it works fine on the Yale cluster now.

psurukuc (approved):

PR #167, opened 2020-10-07
Feature/streamlineparameters

pslocum:
This is underway but is not ready to be merged yet.


pslocum:
Empty config brackets in json file are no longer needed.  Next steps:  Recover functionality to define parameters on command-line.

PR #171, opened 2020-11-30
Feature/ftg fscd

buzinsky:
I thought this would be low effort & with people actively racing to figure out DBF with grad-B motion, I figured there would be a strong desire to have something decent now, as opposed to something with more features in a month or two.

The code does compile and run fine on my machine (mac). Signals seem to be sensible (below). Axial effects, pitch angle effects, and scattering not included (I put in a hardcoded 10 eV energy loss on scatters). For the non-90 pitch angles, I don't think analysis developed enough where a FTG is worthwhile. We of course can move scattering in, I think its not the immediate feature people are looking at, so its worth stepping back and seeing if there's a modular way to include that scattering code.

**Tests**
Plot of signals from channels 1, 16 in 30 channel array, electron at center (expected from spiral)
opposite_phases

Plot of signals from channels 1, 8 in 30 channel array, electron at r=2cm, phi=0 rad.
Figure_1

I think this second plot is the key one, as you can confirm the frequencies are different by looking at where the graphs intersect. The grad-B motion was cranked up by x100 to see the effect in such a short time period

Outside normal review, I have one main request for @psurukuc  : If you run your ellipse producing code on this FTG, do the ellipses match up with your results from the locust LW calculations? Hopefully this isn't too burdensome. If they don't match, my hypothesis that these shifts come from taking the CRES ball and letting normal vector / radius vary is insufficient

If this signal assumption doesn't line up with the more detailed locust simulations, there is still usable stuff here, certain functions would need to be upgraded (not sure how)

Let me know any other thoughts. Thanks

### DIFF
--- a/Config/LocustFreeSpaceFakeTrack.json.in
+++ b/Config/LocustFreeSpaceFakeTrack.json.in
@@ -1,0 +1,73 @@
+{
+    "generators":
+    [          
+       "free-space-fake",
+       "lpf-fft",
+       "decimate-signal",
+       "gaussian-noise",
+       "digitizer"
+    ],
+
+    "free-space-fake":
+    {
+    "start-frequency":
+        {
+            "name": "dirac",
+            "value": 26.025e9
+        },
+    "slope":
+        {
+            "name":"dirac",
+            "value": 0.0
+        },
+    "track-length":
+        {
+            "name":"dirac",
+            "value": 4.0e-6
+        },
+    "radius":
+        {
+            "name":"dirac",
+            "value": 0.02
+        },
+    "grad-B-frequency": 2.410e3,
+    "signal-power": 1.0e-15,
+    "start-vphase": 0.0,
+    "start-time-min": 0,
+    "start-time-max": 1e-7,
+    "lo-frequency": 26.0e9,
+    "ntracks-mean": 1.0,
+    "random-seed": 17,
+    "n-events": 1,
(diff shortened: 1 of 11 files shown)


psurukuc:
@buzinsky Thanks for the quick turnaround on this work. I can surely compare the FTG results to my results. I have never actually run FTG before, could you please provide me with an example config file so I can make the comparison.

pslocum (approved):
It looks ready to merge pending the consistency check.

buzinsky:
[FakeIII.txt](https://github.com/project8/locust_mc/files/5623561/FakeIII.txt)

Just change the extension back to .json & change the paths inside. I also messed with the record-size, so its maybe smaller than you are used to

buzinsky:
Figure_1

okay, think this is pretty nearly shows that we have identified the right feature from our LW solutions causing this effect & there are no egregious errors in this code. 
After this latest fix, electron is at (2cm,0). Seems like the grad-B doppler is dominant here. 
We may want to be super careful & coordinate to show that the magnitude of these effects match with your plots before merging (eg I would need your grad-B frequency as input: I know I put mine absurdly large). 

My initial plots are bugged, exaggerating the frequency differences. If you guys have more comments on the code or checks to the code let's wrap those up here: we should also coordinate what we talk about on the analysis call, which can be moved to slack

psurukuc (approved):

PR #175, opened 2021-01-12
using v_phase instead of c for calculating k_lambda

cclaessens:
for more accurate power dependence on pitch angle


pslocum (approved):
This compiles and runs on my laptop.  

Does it require a new release?

pslocum:
This update looks like it should be in the container on the Yale cluster.  I am going to merge it and do a patch release of Locust for that purpose, probably Wednesday morning, unless there are going to be any other updates?

PR #179, opened 2021-02-08
Feature/ftg aharmonic

buzinsky:
Tested, seems to work
Requested from Spectrum Analysis group

Produces about ~1e-6 level corrections for 0.1 deg. pitch angle changes or so (so kHz)
putting pitch angle dependence of freq. shifts for the non-harmonic traps

Tested on latest stable release of Ubuntu (long story)

To @cclaessens Let me know if you will want more control beyond turning off and on, or if the interface makes sense

### DIFF
--- a/Source/Generators/LMCFakeTrackSignalGenerator.cc
+++ b/Source/Generators/LMCFakeTrackSignalGenerator.cc
@@ -658,6 +658,19 @@ namespace locust
         return j0(k_lambda * zMax);
     }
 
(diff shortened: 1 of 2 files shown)


pslocum (approved):
This builds and runs on my ubuntu v18 and in the p8compute container.

pslocum:
Please let me know if this is ready to go into a new release.

pslocum (approved):

PR #182, opened 2021-02-18
Feature/exit codes

pslocum:
Integer exceptions are thrown for particular failure modes.  The idea is to have the exit codes readable by either e.g. Slurm or Python, or something else, so that the job can be identified and resubmitted if needed.  I am thinking to merge this after we confirm that it is what we want, and that it can work in the script(s).

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ After running Katydid, the processed fft spectra will have a span from 0 to the 
 
 ### Tutorial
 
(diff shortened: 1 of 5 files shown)


MCFlowMace (approved):
That should definitely work for a python script. Checking these exit codes is straight forward, I just checked that.

psurukuc (approved):
Looks good to me

pslocum:
ok sounds good.  Other integer exit codes can also be added as needed.  Assuming cluster tests (today) go as planned we can merge this.

Issue #85, opened 2019-09-10
Streamline Scarab v2 parameter handling in Locust


pslocum:
In the Locust config files, it would be good not to have to define empty code blocks for generators without parameters.  Also, changing parameters on the command line as with Scarab v1 is possibly not working with Scarab v2.

pslocum:
Commandline parameter specification is now supported using Scarab v2 syntax, as in https://github.com/project8/locust_mc/commit/7f3bc1fb5baba5ed3ed5433b1e5a729305ebf217 .  Syntax for Scarab v1 is still supported for compatibility with existing scripts.

PR #189, opened 2021-04-13
Finish upgrade from Scarab v1 to v2.

pslocum:
When we upgraded from Scarab v1 to v2, we lost the ability to change parameters on the command line.  There are also deprecated Scarab functions still being used in Locust.  The purpose of this pull request is to replace deprecated v1 functions with v2 functions, and try to recover the ability to set parameters on the command line as in https://github.com/project8/locust_mc/issues/85 .


pslocum:
The ability to edit parameters on the command line is restored.  However the syntax to flag the name of the config file is "-c" or "--config", and not "config=" as in Scarab v1.  Do we need backward compatibility to preserve the "config=" syntax?

PR #193, opened 2021-05-17
Feature/allow long fir

pslocum:
These changes include an option "allow-fast-sampling" (default=false) to use fast sampling relative to FIR duration.  In 30 GHz sampling tests using the analytic plane-wave transmitter, the output frequency looks correct, but there is some loss in detected power.  This is likely due to a loss of smooth phase variation between samples inside the FIR convolution.    

If we want to try a high-fidelity fast sampling test in the FSCD, then we probably need a shorter duration FIR to fit the short sampling interval.  On the other hand, if we are only checking output frequencies and features (and not necessarily absolute detected power), then these changes might be sufficient.

### DIFF
--- a/Source/Generators/LMCArraySignalGenerator.cc
+++ b/Source/Generators/LMCArraySignalGenerator.cc
@@ -614,7 +614,7 @@ namespace locust
 
         int nFilterBins = fTFReceiverHandler.GetFilterSize();
         double dtFilter = fTFReceiverHandler.GetFilterResolution();
-        int nFilterBinsRequired = 1. / (fAcquisitionRate*1.e6*aSignal->DecimationFactor()) / dtFilter;
+        int nFilterBinsRequired = std::min( 1. / (fAcquisitionRate*1.e6*aSignal->DecimationFactor()) / dtFilter, (double)nFilterBins );
         if (!fAllowFastSampling) nFilterBinsRequired = nFilterBins;
         unsigned nFieldBufferBins = fFieldBufferSize;
         InitializeBuffers(nFilterBins, nFieldBufferBins);
--- a/Source/Generators/LMCArraySignalGenerator.cc
+++ b/Source/Generators/LMCArraySignalGenerator.cc
@@ -389,10 +389,11 @@ namespace locust
 
 
 
-    double ArraySignalGenerator::GetFIRSample(int nfilterbins, double dtfilter, unsigned channel, unsigned element)
+    double ArraySignalGenerator::GetFIRSample(int nFilterBinsRequired, double dtFilter, unsigned channel, unsigned element)
     {
 
     	double fieldfrequency = EFrequencyBuffer[channel*fNElementsPerStrip+element].front();
+//    	printf("field frequency is %g\n", fieldfrequency); getchar();
     	double HilbertMag = 0.;
     	double HilbertPhase = 0.;
     	double convolution = 0.0;
@@ -405,9 +406,9 @@ namespace locust
     		HilbertMag = HilbertMagPhaseMean[0];
     		HilbertPhase = HilbertMagPhaseMean[1];
 
-    		for (int i=0; i < nfilterbins; i++)  // populate filter with field.
+    		for (int i=0; i < nFilterBinsRequired; i++)  // populate filter with field.
     		{
-    			HilbertPhase += 2.*3.1415926*fieldfrequency*dtfilter;
+    			HilbertPhase += 2.*3.1415926*fieldfrequency*dtFilter;
     			ElementFIRBuffer[channel*fNElementsPerStrip+element].push_back(HilbertMag*cos(HilbertPhase));
     			ElementFIRBuffer[channel*fNElementsPerStrip+element].pop_front();
     		}
@@ -421,7 +422,7 @@ namespace locust
     }
 
 
-    bool ArraySignalGenerator::DriveAntenna(FILE *fp, int startingIndex, unsigned index, Signal* aSignal, int nfilterbins, double dtfilter)
+    bool ArraySignalGenerator::DriveAntenna(FILE *fp, int startingIndex, unsigned index, Signal* aSignal, int nFilterBinsRequired, double dtFilter)
     {
 
         const int signalSize = aSignal->TimeSize();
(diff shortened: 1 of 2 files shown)


psurukuc:
Hi Penny, thanks for the edit. Looks like this modification would necessarily just make the ElementFIRBuffer used in convolution longer ? For a full convolution with proper sampling, we will probably need to make the FIR binwidth finer. I still don't understand why the power is lower. Given that the only change this would make is make ElementFIRBuffer I assume the FIR still contains all the needed information. I don't completely understand what you mean by "shorter duration FIR".

pslocum:
Thanks for taking a look at this and for chatting.  I think we can enforce
the smoother phase variation along with faster sampling, and then see how
things are looking at that point.

pslocum:
With "allow-fast-sampling"=true for zero modulation, we now have the same power in the carrier as for "allow-fast-sampling"=false (plots below).  This improvement comes from enforcing smooth phase advancement of the incident field inside the FIR convolution.  For fields with moderate modulation (Trap 10 pitch=88.5, below), we are seeing slightly more power in sidebands and slightly less in the carrier for "allow-fast-sampling"=true.  

We may want to make "allow-fast-sampling" to default to true.  I think it is a better representation of the modulating incident fields.  

newsampling_PlaneWave
oldsampling_PlaneWave
newsampling_Trap10_pitch88 5
oldsampling_Trap10_pitch88 5

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -394,7 +395,7 @@ namespace locust
 
 
 
-    double ArraySignalGenerator::GetFIRSample(int nFilterBinsRequired, double dtFilter, unsigned channel, unsigned element)
+    double ArraySignalGenerator::GetFIRSample(int nFilterBins, int nFilterBinsRequired, double dtFilter, unsigned channel, unsigned element)
```
psurukuc, line 398: I think nFilterSize doesn't need to passed as an argument anymore to GetFIRSample

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -426,7 +444,7 @@ namespace locust
     }
 
 
-    bool ArraySignalGenerator::DriveAntenna(FILE *fp, int startingIndex, unsigned index, Signal* aSignal, int nFilterBinsRequired, double dtFilter)
+    bool ArraySignalGenerator::DriveAntenna(FILE *fp, int startingIndex, unsigned index, Signal* aSignal, int nFilterBins, int nFilterBinsRequired, double dtFilter)
```
psurukuc, line 447: Same here

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -464,7 +482,7 @@ namespace locust
                 if (fTextFileWriting==1) RecordIncidentFields(fp,  fInterface->fTOld, elementIndex, currentElement->GetPosition().GetZ(), tFieldSolution[1]);
 
  	            FillBuffers(aSignal, tFieldSolution[1], tFieldSolution[0], fphiLO, index, channelIndex, elementIndex);
- 	            double VoltageFIRSample = GetFIRSample(nFilterBinsRequired, dtFilter, channelIndex, elementIndex);
+ 	            double VoltageFIRSample = GetFIRSample(nFilterBins, nFilterBinsRequired, dtFilter, channelIndex, elementIndex);
```
psurukuc, line 485: Same here

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -605,7 +625,7 @@ namespace locust
         {
         	for( unsigned index = 0; index < aSignal->DecimationFactor()*aSignal->TimeSize(); ++index )
         	{
-        		DriveAntenna(fp, PreEventCounter, index, aSignal, nFilterBinsRequired, dtFilter);
+        		DriveAntenna(fp, PreEventCounter, index, aSignal, nFilterBins, nFilterBinsRequired, dtFilter);
```
psurukuc, line 628: Same as above

Source/Generators/LMCArraySignalGenerator.hh
```diff
@@ -94,7 +94,7 @@ namespace locust
 
         	void InitializeFieldPoints(std::vector< Channel<Receiver*> > allRxChannels);
             void RecordIncidentFields(FILE *fp, double t_old, int patchIndex, double zpatch, double tEFieldCoPol);
-            double GetFIRSample(int nfilterbins, double dtfilter, unsigned channel, unsigned patch);
+            double GetFIRSample(int nFilterBins, int nFilterBinsRequired, double dtfilter, unsigned channel, unsigned patch);
```
psurukuc, line 97: Same here

Source/Generators/LMCArraySignalGenerator.hh
```diff
@@ -108,10 +108,11 @@ namespace locust
             std::vector<std::deque<double>> LOPhaseBuffer;
             std::vector<std::deque<unsigned>> IndexBuffer;
             std::vector<std::deque<double>> ElementFIRBuffer;
+            std::vector<std::deque<double>> FIRfrequencyBuffer;
 
 
             bool DoGenerate( Signal* aSignal );
-            bool DriveAntenna(FILE *fp, int startingIndex, unsigned index, Signal* aSignal, int nfilterbins, double dtfilter);
+            bool DriveAntenna(FILE *fp, int startingIndex, unsigned index, Signal* aSignal, int nFilterBins, int nFilterBinsRequired, double dtfilter);
```
psurukuc, line 115: Same here

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -656,7 +676,7 @@ namespace locust
                         fInterface->fDigitizerCondition.wait( tLock );
                         if (fInterface->fEventInProgress)
                         {
-                    		if (DriveAntenna(fp, startingIndex, index, aSignal, nFilterBinsRequired, dtFilter))
+                    		if (DriveAntenna(fp, startingIndex, index, aSignal, nFilterBins, nFilterBinsRequired, dtFilter))
```
psurukuc, line 679: Here as well

psurukuc (changes requested):
Looks good Penny. Just for cleaner code you may want to remove "nFilterBins" as argument for GetFIRSample and DriveAntenna functions.

psurukuc (approved):
Looks good to me penny, thanks for doing this.

PR #200, opened 2021-09-23
Feature/vtk support

nsoblath:


pslocum:
There may be some overlap between this branch and https://github.com/project8/locust_mc/pull/198 , now closed.  With the upgrade to Kass v3.8.0, Locust is now compiled with VTK.  Should we still keep this pull request open, try to merge, or possibly close?  Thanks.

PR #219, opened 2022-01-12
Tukey window had an error

arinabykadorova:
The calculation of the tukey window had an error. I don't think it affected anything up until now because it's off in the 25 GHz and 27 GHz regions. However, it did show up when I did wider frequency sweeps. 

The main issue was that the tukey window calculation is calculated using the sample number in the window. See [the wikipedia page here](https://en.wikipedia.org/wiki/Window_function#Tukey_window). However, our window sits inside of a zero-padded array of much longer length. Up until now, we had been using the index inside of that longer array, causing errors in the calculation of the window. At the default parameter alpha = 0.5 and the default n-shift-bins, these differences were minor, but if you move away from 0.5 then the window becomes completely wrong.

In this hotfix, I introduced a new index into the introduced to keep track of the window itself, rather than the larger zero-padded array that it sits inside of. I will be releasing a feature branch with expanded window capabilities soon, but I figured this should be merged in quickly. Also, I fixed some of the formatting, which is why so many lines are changed in the commit.

Here is a plot of the tukey window before and after the hotfix. Like I said, it's not a very big difference for alpha = 0.5 and a suitable choice of n-shift-bins, but for other values of those parameters it would have a big effect on this plot. The sinusoidal sides of the window wouldn't match up between the floor and the flat top of the window, being in the wrong phase.

image
image

And this plot shows the difference before and after the hotfix, but with a frequency sweep using the new port TF. The oscillating parts are due to the hilbert transform not having a window function. I'll be addressing that in my next feature branch.
image

### DIFF
--- a/Source/Transforms/LMCComplexFFT.cc
+++ b/Source/Transforms/LMCComplexFFT.cc
@@ -144,6 +144,7 @@ namespace locust
                 else if(i>midWindowFinalBin && i<=finalWindowFinalBin)
                 {
                     fWindowFunction[i]=0.5*(1 - std::cos( (LMCConst::Pi()*2*(fSize-j)) / (tukeyWindowAlpha*fSize) ) );
+                    j++;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


psurukuc (approved):

pslocum (approved):
This looks good to me.

PR #222, opened 2022-01-24
Feature/equivalent circuit

jkgaison65:
Adding features to configure cavity as an effective circuit model. Interfaces via Transfer Function (TF) and Finite Impulse Response (FIR) related classes implemented previously in Locust.


Source/Core/LMCHFSSResponseFileHandler.cc
```diff
@@ -108,14 +114,24 @@ namespace locust
             fTFComplex[i][0]=tfArray.at(i).real();
             fTFComplex[i][1]=tfArray.at(i).imag();
         }
-        fComplexFFT.SetupIFFT(fTFNBins,fInitialTFIndex,fTFBinWidth);
-	fFIRNBins=fTFNBins+2*fComplexFFT.GetShiftNBins();
+	double TFBinWidth = 2./9.*fInitialTFIndex/(1.0*fTFNBins); //Calculates TFBinWidth assuming frequency range defined as .9 to 1.1 the center of the frequency bins.
```
arinabykadorova, line 117: I don't really understand the explanation in the comment. Could you make it a bit clearer?

Source/Core/LMCHFSSResponseFileHandler.cc
```diff
@@ -108,14 +114,24 @@ namespace locust
...
+	if(GeneratedTF){ 
+		fComplexFFT.SetupIFFT(fTFNBins,fInitialTFIndex,TFBinWidth);//Uses binwidth as calculated in the previous line based on internally generated TF
+	}
+	else{ 
+		fComplexFFT.SetupIFFT(fTFNBins,fInitialTFIndex,fTFBinWidth); //Uses degenerately defined fTFNBins AND fTFBinWidth if using external TF function.
+	}
+	fFIRNBins=fTFNBins+2*2*fComplexFFT.GetShiftNBins();
```
arinabykadorova, line 124: Could this also be put inside of the if-statements above? Where the analytically generated FIR Bins is longer, but the numerical FIR is shorter? I think this is a bit too long for the HFSS TFs.

arinabykadorova (changes requested):
I left a couple of comments in LMCHFSSResponseFileHandler.cc. They're pretty small, but I think would be best addressed before a merge.

Source/Core/LMCHFSSResponseFileHandler.cc
jkgaison65: While I definitely have a typo of an extra factor of 2 on the `GetShiftNBins()` term, are you suggesting that this term should be defined differently for each of the two cases (analytically generated TF vs external TF)? As of now, they both use the same two parameters from the config files. As of now only the bin width is defined differently (this is done to preserve some backwards compatibility of the code), but that isn't used to calculate `fFIRNBins`.

arinabykadorova: Sorry - I thought the typo was intentional and necessary for the generated TF. If you just get rid of the extra 2, then this is resolved.

Config/LocustCavityEqCircuitTest.json
```diff
@@ -0,0 +1,48 @@
...
+       "cavity-signal",
+       "lpf-fft",
+       "decimate-signal",
+       "digitizer"
+    ],
+
+    "cavity-signal":
+    {
+        "transmitter": "kass-current",
+	"e-gun": false,
```
psurukuc, line 13: Looks good in general, but would suggest to maintain same level of indentation for readability.

Config/LocustEquivalentCavity.json
```diff
@@ -0,0 +1,45 @@
+{
```
psurukuc, line 1: What is the difference b/w this file and LocustCavityEqCircuitTest.json ? 
If they are the same, could remove one of them to avoid confusion in the future.

Source/Core/LMCHFSSResponseFileHandler.cc
psurukuc: Why not just assign `2./9.*fInitialTFIndex/(1.0*fTFNBins)` to `fTFBinWidth` directly. So something like for line 117:
```if(GeneratedTF)
{
     double fTFBinWidth = 2./9.*fInitialTFIndex/(1.0*fTFNBins);
}
```
And then you don't need the if-else condition here and you can just do
``` 
fComplexFFT.SetupIFFT(fTFNBins,fInitialTFIndex,fTFBinWidth);
```

Source/Core/LMCHFSSResponseFileHandler.cc
```diff
@@ -180,12 +196,35 @@ namespace locust
...
+    {
+
+	//Replaces ReadHFSSFile() in the case where a Transfer Funciton has been generated analytically
+
+        if(fIsFIRCreated)
+        {
+            return true;
+        }
+        fTFNBins=0;
+        double tfIndex;
```
psurukuc, line 216: What are tfIndex, tfRealValue, and tfImaginaryValue used for  ? If they are not uses, you could remove them.

Source/Core/LMCHFSSResponseFileHandler.cc
psurukuc: In fact why not do define `fTFBinWidth` in `GenerateAnalyticTFtoFIR` right after you define `fTFNBins` ? You won't need these additional if loops at all.

Source/Core/LMCHFSSResponseFileHandler.hh
```diff
@@ -73,14 +74,15 @@ namespace locust
         // Member functions
         virtual bool Configure( const scarab::param_node& aNode) override;
         bool ReadHFSSFile() override;
+	bool GenerateAnalyticTFtoFIR(double initialFreq, std::vector<std::complex<double>> tfArray) override;
```
psurukuc, line 77: Would it make sense to call this `ConvertAnalyticTFtoFIR` ? 
You could also just move all the functionality you define in this function into an ` if(GeneratedTF)` loop inside the 
`ConvertTFtoFIR` and that way you can completely avoid having this function and just directly call `ConvertTFtoFIR` function.

Source/RxComponents/LMCEquivalentCircuit.hh
```diff
@@ -0,0 +1,47 @@
...
+
+
+namespace locust
+{
+ /*!
+ @class LMCEquivalentCicuit
+ @author J. Gaison
+ @brief Base class to implement cavity parameterization
+ @details
+ Available configuration options:
```
psurukuc, line 27: You can remove these lines about available configuration options; looks like they are all defined in LMCCavitySignalGenerator.

psurukuc (changes requested):
Looks good overall, there are a few comments I left.

Source/Core/LMCHFSSResponseFileHandler.cc
jkgaison65: I realized that it was a little confusing having both a `TFBinWidth` and `fTFBinWidth` variable, so I renamed and moved one to `AnalyticTFBinWidth` to more clearly show the differentiate them. I've also updated the comments to (hopefully) be more clear.

	if(GeneratedTF){ 
		//if TF generated based on config file (frequency ranges from 0.9 - 1.1 times the center given in .json config file), calculate the TF bin width given number of bins (also set in .json config file)).
		double AnalyticTFBinWidth = 2./9.*fInitialTFIndex/(1.0*fTFNBins);
		fComplexFFT.SetupIFFT(fTFNBins,fInitialTFIndex,AnalyticTFBinWidth);//Uses binwidth as calculated in the previous line based on internally generated TF
	}
	else{ 
		//if TF read from externally generated TF file, use TF bin width as given in .json config file
		fComplexFFT.SetupIFFT(fTFNBins,fInitialTFIndex,fTFBinWidth); //Uses degenerately defined fTFNBins AND fTFBinWidth if using external TF function.
	}

Source/Core/LMCHFSSResponseFileHandler.cc
jkgaison65: The tricky part here is that I'm attempting to keep this framework flexible so that it still works if being run the original way for when users don't want to use the equivalent circuit model. The original way defines the range of the frequencies explicitly in the config file via the number of bins as well as the bin width (where the initial frequency is taken directly from the external TF file). With the loops, this allows the FFT to either be setup using the bin width as calculated from the internally generated TF function or using the parameters defined in the earlier method. There certainly may be a cleaner way to do this, but I wanted to try to leave the original method as untouched as possible.

Config/LocustCavityEqCircuitTest.json
jkgaison65: Sorry about that, I think my editor has a differently defined tab than the original author. I believe I've fixed it here to be consistent now.

Config/LocustEquivalentCavity.json
jkgaison65: You are correct that these are identical. I think I just used `cp` rather than `mv` when I renamed this file. I've since removed one of the versions to avoid confusion.

Source/Core/LMCHFSSResponseFileHandler.hh
```diff
@@ -26,6 +26,7 @@ namespace locust
         // Member functions
         virtual bool Configure( const scarab::param_node& aNode);
         virtual bool ReadHFSSFile();
+	virtual bool GenerateAnalyticTFtoFIR(double initialFreq, std::vector<std::complex<double>> tfArray);
```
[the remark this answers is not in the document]
jkgaison65: You are correct regarding the double definition, so I've removed the original instance of it. I tried to match the structure of the original method that reads in the TF file directly (defined through `ReadHFSSFile()` in this class), which is why it's defined here, although it could be defined in TFFileHandlerCore directly.

Source/Generators/LMCCavitySignalGenerator.cc
```diff
@@ -249,10 +249,61 @@ namespace locust
...
-        {
-            return false;
-        }
+
+//-------------New implementation of fEquivalentCircuit for cavity parameterization-----
+	if ( (aParam.has( "equivalentR" )) or (aParam.has( "equivalentL" )) or (aParam.has( "equivalentL" )) ) { //Only initiate the configurable TF if at least one parameter is specified in the config file
+		//Default RLC parameters if not defined in config file, values (mostly arbitratily) based on external script from P. Slocum for 1 GHz cavity
+		printf("Entering RLC Config Loop\n");
+		double equivalentR = 1.;
+                double equivalentL = 0.159e-6;
```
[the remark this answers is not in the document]
jkgaison65: Similar to your previous comment, I think this deals with different default tab settings. I believe it is consistent with the rest of the file now.

Source/Core/LMCHFSSResponseFileHandler.cc
jkgaison65: I believe my response earlier addresses most of the comments besides the last one. You are correct that those three objects aren't used here (they're leftovers from an earlier version of the code) and have been removed.

Source/Core/LMCHFSSResponseFileHandler.hh
jkgaison65: I agree that `ConvertAnalyticTFtoFIR` is more intuitive, so I have changed it.

Source/RxComponents/LMCEquivalentCircuit.hh
jkgaison65: These lines have been removed for your given reasons.

psurukuc (approved):

Source/Core/LMCHFSSResponseFileHandler.cc
psurukuc: I don't understand the logic here. 
Why is the `bin width = 2/(9*initialindex*nBins)` ?

psurukuc:
Looks good Jeremy, thanks for addressing my comments. I only have one comment about logic, other than that I approved the PR.

arinabykadorova (approved):

pslocum:
Thanks for all of the comments and suggestions.  This new functionality will be very useful.  Also, I confirmed backward-compatibility with existing work.

PR #223, opened 2022-01-26
Feature/cavity development

pslocum:
These commits contain supporting steps toward normalizing the cavity simulation.  The merge is being requested now to keep up with related development on other branches.  
modes

### DIFF
--- a/Source/Fields/LMCCylindricalCavity.cc
+++ b/Source/Fields/LMCCylindricalCavity.cc
@@ -77,25 +77,33 @@ namespace locust
     	return tIntegral;
     }
 
-    double CylindricalCavity::Z_TE(int l, int m, int n) const
+    double CylindricalCavity::Z_TE(int l, int m, int n, double fcyc) const
     {
+    	double Z_TE = 1.0;
     	double x_lm = fInterface->fBesselNKPrimeZeros[l][m];
     	double k1 = x_lm / fInterface->fR;
     	double k3 = n * LMCConst::Pi() / fInterface->fL;
     	double k = pow(k1*k1+k3*k3,0.5);
-    	double eta = sqrt( LMCConst::MuNull() / LMCConst::EpsNull() );  // Pozar p. 291.
-    	double Z_TE = k * eta / k3; // Pozar p. 286, Jackson Eq. 8.32
+    	double k0 = fcyc / LMCConst::C();
+    	if ( k*k-k0*k0 != 0. )
+    	{
+    		Z_TE *= fcyc/(k0*k0 - k*k);  // after Collin Foundations of M.E. Eq. 7.132
+    	}
     	return Z_TE;
     }
 
-    double CylindricalCavity::Z_TM(int l, int m, int n) const
+    double CylindricalCavity::Z_TM(int l, int m, int n, double fcyc) const
     {
+    	double Z_TM = 1.0;
     	double x_lm = fInterface->fBesselNKZeros[l][m];
     	double k1 = x_lm / fInterface->fR;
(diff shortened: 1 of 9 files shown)


psurukuc (approved):
I don't understand all the logic, but generally looks good Penny.

Source/Fields/LMCCylindricalCavity.cc
```diff
@@ -77,25 +77,33 @@ namespace locust
...
     	double k1 = x_lm / fInterface->fR;
     	double k3 = n * LMCConst::Pi() / fInterface->fL;
     	double k = pow(k1*k1+k3*k3,0.5);
-    	double eta = sqrt( LMCConst::MuNull() / LMCConst::EpsNull() );  // Pozar p. 291.
-    	double Z_TE = k * eta / k3; // Pozar p. 286, Jackson Eq. 8.32
+    	double k0 = fcyc / LMCConst::C();
+    	if ( k*k-k0*k0 != 0. )
+    	{
+    		Z_TE *= fcyc/(k0*k0 - k*k);  // after Collin Foundations of M.E. Eq. 7.132
+    	}
```
jkgaison65, line 91: I know this is still a work in progress, but do these impedance factors have other components that will be incorporated later? Perhaps I'm misreading the referenced equation but it seems like there are more components to the given term.

Source/Generators/LMCCavitySignalGenerator.cc
```diff
@@ -531,50 +553,63 @@ namespace locust
...
+    						// TO-DO:  Implement modeScalingFactor to calculate signal amplitude at probe/channel using
+    						// H-field and probe inductance.  Probe inductance is presently set to a fake value of 1.0
+    						// in LMCPowerCombiner.cc
+    						// double modeScalingFactor = GetModeScalingFactor(tKassParticleXP, channelIndex);
+    						double modeScalingFactor = 1.0; // override for now.
+    						// end TO-DO
+
+    						// This scaling factor includes a 50 ohm impedance that is expected in signal processing, as well
+    						// as other factors as defined above.
+    						double totalScalingFactor = sqrt(50.) * unitConversion * impedanceFactor * modeScalingFactor * dotProductFactor * modeAmplitude;
```
jkgaison65, line 602: Again I know many of these factors are placeholders for future work, but it isn't intuitive to me that there are separate factors for the 50 ohm impedance as well as an `impedanceFactor`. Would it make sense to rename the latter to something like `expansionCoefficient` if it's based off of the equations from Foundations for ME? I think the current variable name comes from the older version of the code where you were using the Jackson definition of an actual mode impedance term.

jkgaison65 (approved):
Structurally I think this all looks great. I just had a couple comments regarding some of the modified terms in the math, but most of it might just be your own placeholders for future updates.

Source/Fields/LMCCylindricalCavity.cc
pslocum: Thanks and yes, it does need a closer look in the context of the other steps in the simulation.  So this first step only contains the frequency and wavenumbers as a first pass.

Source/Generators/LMCCavitySignalGenerator.cc
pslocum: The 50 ohms is the receiver impedance assumed and applied in katydid post processing.  So I think it is distinct from the field coefficients here.  In the meantime I agree we should rename impedanceFactor to be expansionCoefficient.  This change has been committed.

pslocum:
Thanks again for the comments.  I think we can merge this now so that the next steps can include new work from other branches.

PR #225, opened 2022-02-01
Window functions implemented into FFTs

arinabykadorova:
This is a relatively large-scale effort that implements a new window function class, and integrates it into the generation of FIRs and into the Hilbert transform. Now, there is always a window function applied explicitly before an FFT. This is configurable. Currently the options are rectangular, tukey, and hann windows. The defaults have been changed such that running without specifying configurations will work reasonably.
Major changes for review are in LMCComplexFFT, LMCHFSSResponseFileHandler, and LMCHilbertTransform, as well as the new LMCWindowFunctions class.


Source/RxComponents/LMCPowerCombiner.cc
```diff
@@ -18,19 +18,19 @@ namespace locust
     PowerCombiner::PowerCombiner():
 			fnElementsPerStrip( 0 ),
 			fdampingFactors( 0 ),
-            fjunctionLoss( 1.0 ),
-            fpatchLoss( 0.6 ),
-            famplifierLoss( 0.66 ),
-            fendPatchLoss( 1.0 ),
-            fjunctionResistance( 0.3 ),
-            fnCavityProbes( 0 ),
+      fjunctionLoss( 1.0 ),
```
psurukuc, line 21: I suggest that we maintain consistent indentation atleast within a file. I have had a lot of trouble reading through code in the past when not indented properly.

arinabykadorova: Sorry. I think there was a conflict here between my editor's indentations and Penny's.
I changed a lot of indentation in ComplexFFT, which seems to have survived. I'll fix this one, too.

pslocum (commented):
Thank you very much for this pull request.  I would like to take a closer look at the modifications, with a goal to identify a straightforward path for people to reproduce plots made before the modifications.  I will try to do this asap.

Source/Transforms/LMCWindowFunctions.hh
```diff
@@ -0,0 +1,56 @@
...
+    class WindowFunctions
+    {
+        
+    public:
+        WindowFunctions();
+        virtual ~WindowFunctions();
+        bool Configure( const scarab::param_node& aNode);
+        bool GenerateWindowFunction(int windowsize, int prewindowbins, int signalsize);
+        bool SetupWindow(std::string windowname, double windowparam);
+        std::vector<double>* GetWindowFunction();
```
psurukuc, line 34: Here are a couple of changes to return vector the vector is an slighlty cleaner way. I will put comments at other places to account for this change. 
Here, I'd change this to 
```const std::vector<double>* GetWindowFunction() const;```
This will return a vector that can't be modified outside of this class.

File: Source/Transforms/LMCWindowFunctions.cc [hunk 3]
```diff
@@ -0,0 +1,172 @@
...
+            {
+                fWindowFunction[i]=0.5*(1 - std::cos( (2*LMCConst::Pi()*j) / fSignalSize ) );
+                j++;
+            }
+        }
+    }
+
+    std::vector<double>* WindowFunctions::GetWindowFunction()
+    {
+        return fWindowPtr;
```
psurukuc, line 164: You can just do 
```return &fWindowFunction``` here and remove the need to have a separate pointer `fWindowPtr`.

File: Source/Transforms/LMCWindowFunctions.cc [hunk 4]
```diff
@@ -0,0 +1,172 @@
...
+        else if(fWindowFunctionType==2)
+        {
+            GenerateHanningWindow();
+        }	
+        else
+        {
+            return false;
+        }
+
+        fWindowPtr = &fWindowFunction;
```
psurukuc, line 92: Could remove this line if you are returning the reference to `fWindowFunction` as suggested

Source/Transforms/LMCWindowFunctions.hh
```diff
@@ -0,0 +1,56 @@
...
+
+    private:
+    
+        int fWindowFunctionType;
+        int fTotalWindowSize;
+        int fPreWindowBins;
+        int fSignalSize;
+        double fTukeyWindowAlpha;
+        std::vector<double> fWindowFunction;
+        std::vector<double>* fWindowPtr;
```
psurukuc, line 45: Could remove this line if you are returning the reference to fWindowFunction as suggested

psurukuc (changes requested):

Source/Transforms/LMCHilbertTransform.cc
```diff
@@ -65,6 +100,14 @@ namespace locust
...
     }
 
+    bool HilbertTransform::SetupHilbertTransform()
+    {
+        originaldata = (fftw_complex *) fftw_malloc(sizeof(fftw_complex) * fbufferSize);
+        SignalComplex = (fftw_complex*)fftw_malloc( sizeof(fftw_complex) * fbufferSize );
+        FFTComplex = (fftw_complex*)fftw_malloc( sizeof(fftw_complex) * fbufferSize );
+        hilbert = (fftw_complex *) fftw_malloc(sizeof(fftw_complex) * fbufferSize);
+        return fComplexFFT.SetupFFTWindow(fbufferSize, fWindowName, fWindowParam);
+    }
```
pslocum, line 110: This looks great.  Can it be called from within HilbertTransform::Configure() instead of from LMCArraySignalGenerator?

Source/Transforms/LMCHilbertTransform.cc
```diff
@@ -169,32 +212,22 @@ namespace locust
...
-
-        fftw_complex *originaldata;
-        originaldata = (fftw_complex *) fftw_malloc(sizeof(fftw_complex) * windowsize);
-        fftw_complex *SignalComplex;
-        SignalComplex = (fftw_complex*)fftw_malloc( sizeof(fftw_complex) * windowsize );
-        fftw_complex *FFTComplex;
-        FFTComplex = (fftw_complex*)fftw_malloc( sizeof(fftw_complex) * windowsize );
-        fftw_complex *hilbert;
-        hilbert = (fftw_complex *) fftw_malloc(sizeof(fftw_complex) * windowsize);
+        int windowsize=FieldBuffer.size()-1;
```
pslocum, line 215: Shouldn't the windowsize be the same as FieldBuffer.size()?  What does the `-1`  do here?

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -563,6 +563,12 @@ namespace locust
             return false;
         }
 
+        if(!fHilbertTransform.SetupHilbertTransform())
+        {
+            LERROR(lmclog,"Error initializing Hilbert transform.\n");
+            return false;
+        }
+
```
pslocum, line 571: As mentioned above, can this functionality happen from within HilbertTransform::Configure()?  Or does it need to happen from outside HilbertTransform?

Source/Transforms/LMCHilbertTransform.cc
```diff
@@ -15,13 +15,40 @@ namespace locust
...
     HilbertTransform::HilbertTransform():
-        fbufferMargin( 25 ),
-		fbufferSize( 50 )
+        fbufferMargin( 50 ),
+		fbufferSize( 100 ),
+        originaldata(NULL),
+        SignalComplex(NULL),
+        FFTComplex(NULL),
+        hilbert(NULL),
+        fWindowName( "hanning" ),
```
pslocum, line 24: Can this be changed back to its default value of "rectangular", and instead move this new setting to the default config files in the hexbug repo?

Config/LocustCavityEqCircuitTest.json
```diff
@@ -20,6 +20,8 @@
 	"equivalentL": 0.159e-6,
 	"equivalentC": 0.159e-12,
 	"FreqRangeCenter": 1.0e9,
+    "tf-window-type": "rectangular",
+    "tf-window-parameter": 0.0,
```
pslocum, line 24: Can these lines be removed to preserve defaults as noted below?

pslocum (changes requested):
These look like helpful developments, thank you again for doing this.  With the initial requested changes in place I would next like to compare the output spectra with v2.2.0 for compatibility.

Source/Transforms/LMCHilbertTransform.cc
arinabykadorova: Yes, I think so. I'll move it and check.

Source/Transforms/LMCHilbertTransform.cc
arinabykadorova: It turns out that FieldBuffer.size() returns 1 more than the actual buffer size. So for a buffer size of 50, it returns 51. There is probably an off-by-one error upstream, but I'm not sure where it is.

Source/Generators/LMCArraySignalGenerator.cc
```diff
@@ -42,7 +42,7 @@ namespace locust
         IndexBuffer( 1 ),
         ElementFIRBuffer( 1 ),
         FIRfrequencyBuffer( 1 ),
-        fFieldBufferSize( 50 ),
+        fFieldBufferSize( 100 ),
```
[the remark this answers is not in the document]
arinabykadorova: okay

Source/Transforms/LMCHilbertTransform.cc
```diff
@@ -15,13 +15,40 @@ namespace locust
 
 
     HilbertTransform::HilbertTransform():
-        fbufferMargin( 25 ),
-		fbufferSize( 50 )
+        fbufferMargin( 50 ),
+		fbufferSize( 100 ),
```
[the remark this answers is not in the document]
arinabykadorova: okay

Source/Transforms/LMCHilbertTransform.cc
arinabykadorova: same for the buffer lengths? 50 is just too short

Source/Transforms/LMCHilbertTransform.cc
arinabykadorova: Actually, I think I know why. The buffer is initialized to have 50 zeros. Then it goes through LMCArraySignalGenerator::DriveAntenna() and FillBuffers is called. Then it has 51 entries. Then GetFIRSample is called, so all that is being done on a buffer that is 51 elements long. Only after that is PopBuffers called.
So either we have to initialize to 49, or pop the buffers before getting the FIR sample.

pslocum: Aha, I see.

Source/Transforms/LMCHilbertTransform.cc
pslocum: yes, I would say that the modified user-specific values for buffer length parameters can best be specified in the hexbug config files.

Source/Transforms/LMCWindowFunctions.hh
arinabykadorova: I couldn't get the const to compile ... could you take a look at my changes and see?

Source/Transforms/LMCComplexFFT.cc
```diff
@@ -248,13 +172,52 @@ namespace locust
...
+    {
+        if(!fWindowFunction.IsWindowGenerated())
+        {
+            LERROR(lmclog,"Error applying window function, it has not been generated");
+            exit(-1); 
+        }
+        double windowfactor = 0.0;
+        for (int i = 0; i < size; ++i)
+        { 
+            windowfactor = fWindowFunction.GetWindowFunction()->at(i+fPreFilterBins);
```
psurukuc, line 194: I think you may have to do this here:
```const aWindowFunction=fWindowFunction.GetWindowFunction()->at(i+fPreFilterBins);```
outside the function and then inside the function 
```windowfactor = aWindowFunction.GetWindowFunction()->at(i+fPreFilterBins);```
This will also reduce the calls to `GetWindowFunction` for every `i` anyway.

File: Source/Transforms/LMCWindowFunctions.hh (hunk 2 above)
Comment (psurukuc) [reply Source/Transforms/LMCWindowFunctions.hh:34]: Hi Arina, I added a couple more comments. Let me know if that works.

**Comment by pslocum** [comment]
Related to the above, I have located the two commits on develop where the benchmarking outputs change compared to v2.2.0 .  The two commits contain important changes and they are 4e3b36cd3 and 487a1ca63c .

**Review by pslocum — COMMENTED** [review]
Thanks for making these changes.  A question:  If I revert the order of the buffer fill/pop and revert the windowsize to FieldBuffer.size()-1 as it is on develop, I am not yet able to reproduce the develop output.  Is there something else I can revert in order to reproduce develop, for the purpose of tracking changes?

**Review by psurukuc — COMMENTED** [review]
Looks good to me, thanks for making the changes.

**Review by psurukuc — APPROVED** [review]

File: Source/Transforms/LMCWindowFunctions.cc [hunk 14]
```diff
@@ -0,0 +1,170 @@
...
+    fWindowFunction( 1 ),
+    fTukeyWindowAlpha( 0.5 )
+    {
+    }
+    
+    WindowFunctions::~WindowFunctions()
+    {
+    }
+    
+    bool WindowFunctions::Configure(const scarab::param_node& aParam)
```
jkgaison65, line 33: I'm a little confused why this is here. It isn't configuring anything, so is this just leftovers from previous versions of the code?

jkgaison65 (approved):
Just one minor comment clarifying a Configure function, but overall it looks good to me

pslocum (commented):
We are looking to characterize remaining differences between this branch and develop, as in the side-by-side benchmarking tests below.  Ideally these changes will be described in release notes, in addition to those identified in [4e3b36c](https://github.com/project8/locust_mc/commit/4e3b36cd3f94f5c487c2160c46059f96d5fdc24e) and [487a1ca](https://github.com/project8/locust_mc/commit/487a1ca63cc65577b79517b36ff321efe992839d) .

featureWindowing
develop

Source/Transforms/LMCWindowFunctions.cc
pslocum: For right now it is a useful placeholder that gets run once after Locust is started.  If the WindowFunctions class needs parameterizable class variables in the future, they can easily be located here.

arinabykadorova:
The major difference between this branch and the current develop branch (c31d450) is the buffer length. Before switching the order of Fill and Pop in the ArraySignalGenerator, the effective default buffer length was 51. If this is switched back in LMCArraySignalGenerator.cc and line 115 of LMCHilbertTransform.cc is changed to `return fComplexFFT.SetupFFTWindow(fbufferSize+1, fWindowName, fWindowParam);`, then the resulting spectra are identical.
After merging, this code will make all the buffer sizes used internally consistent with what is specified in the configuration file.

The plots below show this for various conditions. (All were done with the `LocustFreeSpaceTemplate.json` and default configs in the cbuild/config directory).
develop_c31d450
windowbranch_changedtomatch
currentwindowbranch

pslocum (approved):
Thank you for addressing all of the comments in this thread, and for implementing these helpful changes.

PR #227, opened 2022-03-03
Feature/cavity development

pslocum:
These changes include some new parametrizations for ease of use as in Slide 9 of https://3.basecamp.com/3700981/projects/3107037 .  There are also some changes to variable names for alignment with development on other branches.

### DIFF
--- a/Source/Fields/LMCCylindricalCavity.cc
+++ b/Source/Fields/LMCCylindricalCavity.cc
@@ -132,6 +132,7 @@ namespace locust
 //    		Z_TM *= fcyc * LMCConst::MuNull() / (k0*k0 - k*k);  // neglect Q term.
     		Z_TM *= fcyc * LMCConst::MuNull() * sqrt( real*real + imag*imag ); // mag
     	}
+
     	return Z_TM;
     }
 
@@ -170,9 +171,9 @@ namespace locust
     	double jl_of_k1r_by_k1r = 1./(2.*l) * (boost::math::cyl_bessel_j(l-1, k1*r) + boost::math::cyl_bessel_j(l+1, k1*r));
     	double jPrime = 1./2. * boost::math::cyl_bessel_j(l-1, k1*r) - boost::math::cyl_bessel_j(l+1, k1*r);
     	TE_H.push_back(-k3/k1 * jPrime * cos(l*theta) * cos(k3*z));
-    	TE_H.push_back(-l*k3/k1 * jl_of_k1r_by_k1r * sin(l*theta) * cos(k3*z));
     	TE_H.push_back(boost::math::cyl_bessel_j(l, k1*r) * cos(l*theta) * sin(k3*z));
-        return TE_H;
+    	TE_H.push_back(-l*k3/k1 * jl_of_k1r_by_k1r * sin(l*theta) * cos(k3*z));
+        return TE_H; // r, z, theta
     }
 
     std::vector<double> CylindricalCavity::TM_E(int l, int m, int n, double r, double theta, double zKass, double fcyc) const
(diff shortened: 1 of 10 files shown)


Source/Fields/LMCCylindricalCavity.cc
```diff
@@ -87,33 +87,54 @@ namespace locust
...
     	double k = pow(k1*k1+k3*k3,0.5);
     	double k0 = fcyc / LMCConst::C();
+    	double A = 1.;
+    	double B = k*k;
+    	double C = k0*k0;
+    	double Q = 1.;  // Q for suppressed modes.
+    	if ((l==0)&&(m==1)&&(n==1)) Q = 1000.; // Q for lmn = 011.
+    	double denom = (Q*B + Q*C + C) * (Q*B+ Q*C + C) + C*C;
+    	double real = (Q*A * (Q*B + Q*C + C)) / denom;
+    	double imag = Q*A*C / denom;
```
jkgaison65, line 103: I get a slightly different set of terms when I rearrange eq. 7.132 to get this. I _think_ that the minus sign on the k_0^2 was dropped in your version. I get:

    	double denom = (Q*B - Q*C - C) * (Q*B - Q*C - C) + C*C;
    	double real = (Q*A * (Q*B - Q*C - C)) / denom;
    	double imag =  -1. * Q*A*C / denom;

with the same solutions in the TM version of this function. Also admittedly the minus on the `imag` term doesn't matter since it's only used when squared. I also may have messed up my own algebra, so it might be worth double checking this.

Source/Fields/LMCCylindricalCavity.cc
```diff
@@ -152,9 +173,9 @@ namespace locust
     	double jl_of_k1r_by_k1r = 1./(2.*l) * (boost::math::cyl_bessel_j(l-1, k1*r) + boost::math::cyl_bessel_j(l+1, k1*r));
     	double jPrime = 1./2. * boost::math::cyl_bessel_j(l-1, k1*r) - boost::math::cyl_bessel_j(l+1, k1*r);
     	TE_H.push_back(-k3/k1 * jPrime * cos(l*theta) * cos(k3*z));
-    	TE_H.push_back(-l*k3/k1 * jl_of_k1r_by_k1r * sin(l*theta) * cos(k3*z));
     	TE_H.push_back(boost::math::cyl_bessel_j(l, k1*r) * cos(l*theta) * sin(k3*z));
-        return TE_H;
+    	TE_H.push_back(-l*k3/k1 * jl_of_k1r_by_k1r * sin(l*theta) * cos(k3*z));
+        return TE_H; // r, z, theta
```
jkgaison65, line 178: Sorry, is there a specific reason you're changing the ordering of the coordinates here (and in TM_E)? Was there an issue somewhere else that I'm missing? I only see this implemented in print statements like in new line 257 of `LMCCavitySignalGenerator.cc`

jkgaison65 (changes requested):
Overall looks like a clean update, just one question regarding your implementation of Q into Z_TE and Z_TM

Source/Fields/LMCCylindricalCavity.cc
pslocum: Thank you!  I just recalculated and now I am getting the same answer as you.  I pushed the changes to this branch.

Source/Fields/LMCCylindricalCavity.cc
pslocum: Right, this was done to allow efficient access to `r` and `theta` components of fields with e.g. `tEr = anE_normalized.front()` and  `tEtheta = anE_normalized.back()` , whether the `z` component is present or not.  This was found to help with the recently-implemented TM calculations where the `E` field has 3 components instead of 2.  We could probably accomplish this some other way, but this seemed efficient at the time.

jkgaison65 (approved):

PR #229, opened 2022-03-30
Cleanup of EquivalentCircuit Configure() function

jkgaison65:
Here is no new functionality, I'm mostly just cleaning up the Configure procedure to match the style of the rest of the code (moving configurable parameters from LMCHFSSResponseFileHandler into the EquivalentCircuit class directly). Also updating the default cavityTF.txt transfer function to have a finer binwidth in frequency space so that the full ringdown is captured in the FIR (this had already been addressed in some places, this is just unifying the process).


pslocum (commented):
Thanks very much for these changes, they look great.  I am going to suggest a few minor formatting tweaks in a couple of commits that I will push to this branch for your inspection.

Source/Core/LMCHFSSResponseFileHandler.cc
```diff
@@ -132,12 +132,15 @@ namespace locust
     fComplexFFT.ApplyWindowFunction(fTFNBins, fTFComplex);
     fComplexFFT.GenerateFIR(fTFNBins,fTFComplex,fFIRComplex);
     fResolution=fComplexFFT.GetTimeResolution();
-
-    for (int i = 0; i < fFIRNBins; ++i)
+    FILE * fFIRout = fopen("output/FIR.txt", "w");
+    fprintf(fFIRout,"#FIR for cavity simulation (index,coefficient)\n");
```
pslocum, line 136: Can this comment line be made more general?  I think the file `FIR.txt`  will be written for any TF.  Also, I think it is helpful to output this file routinely.

pslocum (approved):

Source/Core/LMCHFSSResponseFileHandler.cc
jkgaison65: Ok, I adjusted that comment line to remove the explicit cavity reference

PR #231, opened 2022-04-20
Feature/run without digitizer

MCFlowMace:


Source/Core/LMCEggWriter.cc
```diff
@@ -106,34 +106,45 @@ namespace locust
...
-            t_signed_vals = a_digitizer->GetADCValuesSigned();
+            t_data_format = a_digitizer->GetADCValuesSigned();
             t_bit_depth = a_digitizer->DigitizerParams().bit_depth;
             t_bits_right_aligned = a_digitizer->DigitizerParams().bits_right_aligned;
-        }
+            t_v_range = a_digitizer->DigitizerParams().v_range;
+            t_v_offset = a_digitizer->DigitizerParams().v_offset;
+            t_dac_gain = a_digitizer->DigitizerParams().dac_gain;
+        } else {
+			LDEBUG( lmclog, "Running without digitizer, data saved in float format");
```
pslocum, line 127: Consider upgrading from LDEBUG to LWARN.  Probably not very important.

pslocum (commented):
These changes look straightforward and produce normal-looking output for default configs.  What I can't find is how to actually test the new feature.  Is there a parameter that I need to type?

MCFlowMace:
> These changes look straightforward and produce normal-looking output for default configs. What I can't find is how to actually test the new feature. Is there a parameter that I need to type?

With the changes you can just omit `digitizer` from the `generators` list. No new keyword was added.

MCFlowMace:
> These changes look straightforward and produce normal-looking output for default configs. What I can't find is how to actually test the new feature. Is there a parameter that I need to type?

With the changes you can just omit `digitizer` from the `generators` list. No new keyword was added.

pslocum:
This seems to run smoothly and I think it looks ready to be merged, unless we should wait for additional changes?

MCFlowMace:
I do not have any additional changes planned at the moment. We can go ahead from my side.

Issue #201, opened 2021-10-08
Update natural constants to latest PDG values


renereimann:
The values of natural constants are from PDG 2006 (https://pdg.lbl.gov/2006/reviews/consrpp.pdf) which are outdated and have significant larger uncertainties compared to up-to date from PDG 2021 (https://pdg.lbl.gov/2021/reviews/rpp2020-rev-phys-constants.pdf).
Also in the mean time the SI unit definition has been changed. This would effect:
https://github.com/project8/locust_mc/blob/master/Source/Core/LMCConst.hh
Kassiopeia uses the same constants and should be updated at the same time (otherwise we are inconsisten and partly outdated).
https://github.com/KATRIN-Experiment/Kassiopeia/blob/main/Kommon/Core/Utility/KConst.h

renereimann:
In Kassiopeia this has been fixed by https://github.com/KATRIN-Experiment/Kassiopeia/pull/55

PR #235, opened 2022-05-20
Update constants according to PDG 2021.

pslocum:
These changes include updates to natural constants according to https://pdg.lbl.gov/2021 .  This addresses https://pdg.lbl.gov/2021 .

### DIFF
--- a/Source/Core/LMCConst.hh
+++ b/Source/Core/LMCConst.hh
@@ -28,33 +28,33 @@ public:
 
     constexpr static double Q()
     {
-        return 1.60217653E-19;
+        return 1.602176634E-19;  // updated from https://pdg.lbl.gov/2021
     } //!< elementary charge  in C(>0)
 
     constexpr static double Hbar()
     {
-        return 1.05457168E-34;
+        return 1.054571817E-34; // updated from https://pdg.lbl.gov/2021
     }//!< hbar in J s-1
 
     constexpr static double HbarC_eV()
     {
-        return 197.326968E-9;
+    	return 197.3269804; // updated from https://pdg.lbl.gov/2021
     }//!<hbar c in m eV.
 
     constexpr static double kB()
     {
-        return 1.3806505E-23;
+        return 1.380649E-23; // updated from https://pdg.lbl.gov/2021
     }//!< Boltzmann constant J/K
 
     constexpr static double kB_eV()
     {
-        return 8.617343E-5;
+        return 8.617333262E-5; // updated from https://pdg.lbl.gov/2021
     }//!< Boltzmann constant eV/K
 
     //EM coupling constants
     constexpr static double EpsNull()
     {
-        return 8.854187817E-12;
+        return 8.8541878128E-12; // updated from https://pdg.lbl.gov/2021
     } //!< epsilon0, Constant of Newtons force.
 
     constexpr static double FourPiEps()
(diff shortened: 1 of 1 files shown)


Source/Core/LMCConst.hh
```diff
@@ -28,33 +28,33 @@ public:
...
     {
-        return 1.60217653E-19;
+        return 1.602176634E-19;  // updated from https://pdg.lbl.gov/2021
     } //!< elementary charge  in C(>0)
 
     constexpr static double Hbar()
     {
-        return 1.05457168E-34;
+        return 1.054571817E-34; // updated from https://pdg.lbl.gov/2021
     }//!< hbar in J s-1
```
renereimann, line 37: the units of hbar are J * s and not J / s

Source/Core/LMCConst.hh
```diff
@@ -28,33 +28,33 @@ public:
...
 
     constexpr static double HbarC_eV()
     {
-        return 197.326968E-9;
+    	return 197.3269804; // updated from https://pdg.lbl.gov/2021
```
renereimann, line 41: if units should be in eV * m then a factor 1e-9 is missing. In PDG they are given as MeV (1e6) * fm (1e-15) thus canceling the 1e6 and 1e-15 results in 1e-9.

Source/Core/LMCConst.hh
```diff
@@ -28,33 +28,33 @@ public:
...
-        return 8.617343E-5;
+        return 8.617333262E-5; // updated from https://pdg.lbl.gov/2021
     }//!< Boltzmann constant eV/K
 
     //EM coupling constants
     constexpr static double EpsNull()
     {
-        return 8.854187817E-12;
+        return 8.8541878128E-12; // updated from https://pdg.lbl.gov/2021
     } //!< epsilon0, Constant of Newtons force.
```
renereimann, line 58: this is not "Constant of Newton force." but "permittivity of free space" or "electric constant" there are units as well. They are "F / m" (Farad per meter)

Source/Core/LMCConst.hh
```diff
@@ -70,12 +70,12 @@ public:
...
     constexpr static double M_el_kg()
     {
-        return 9.1093826E-31;
+        return 9.1093837015E-31; // updated from https://pdg.lbl.gov/2021
     } //!< electron mass in kg
 
     constexpr static double M_el_eV()
     {
-        return 510.998918E3;
+        return 510.998950E3; // updated from https://pdg.lbl.gov/2021
```
renereimann, line 78: I would write it as `510.99895000E3` (with three zeros at the end) as this is the precision with which the quantity is known and reported.

renereimann (changes requested):
L. 67 After the fixing of some constants in the SI units in 2019 Mu_0 is no longer 4pi*1e-7 but 4*pi*1e-7* 1.00000000055 because this value is no longer defined to be exact but derived from other measurements-

L 63. this is not constant of newtons force

All other comments have been submitted inline

This change should happen with also upgrading to the corresponding Kassiopeia version. As constants are also defined in Kassiopeia they should be consistent.

pslocum:
We will plan to link this update to the work being done in https://github.com/project8/locust_mc/pull/198 , and reference the issue here:  https://github.com/project8/locust_mc/issues/201 .

pslocum:
Suggesting to close this pull request and start a new one, with the goal to sync with a corresponding workflow in Kassiopeia.  Changes requested here are saved and can be ported by hand to the new request.

PR #241, opened 2022-07-08
Feature/cavity development

pslocum:
This is a draft of ongoing work to implement a harmonic oscillator as a cavity response function, as suggested by @marsteller .  It is not ready for testing yet, but the approach can be tracked here in combination with other discussion/suggestions.

### DIFF
--- a/Source/RxComponents/LMCAnalyticResponseFunction.cc
+++ b/Source/RxComponents/LMCAnalyticResponseFunction.cc
@@ -15,7 +15,10 @@ namespace locust
 	LOGGER( lmclog, "AnalyticResponseFunction" );
 
 
-    AnalyticResponseFunction::AnalyticResponseFunction()
+    AnalyticResponseFunction::AnalyticResponseFunction():
+	    fGeneratingTF( false ),
+		fInitialFreq( 0. ),
+		fTFarray(0 )
     {}
     AnalyticResponseFunction::~AnalyticResponseFunction() {}
 
@@ -24,6 +27,36 @@ namespace locust
     {
         return true;
     }
+
+    void AnalyticResponseFunction::SetGeneratingTF( bool aFlag )
+    {
+    	fGeneratingTF = aFlag;
+    }
+    bool AnalyticResponseFunction::GetGeneratingTF()
+    {
+    	return fGeneratingTF;
+    }
+    void AnalyticResponseFunction::SetInitialFreq( double aFreq )
+    {
+    	fInitialFreq = aFreq;
+    }
+    double AnalyticResponseFunction::GetInitialFreq()
+    {
+    	return fInitialFreq;
+    }
+    void AnalyticResponseFunction::SetTFarray( std::vector<std::complex<double>> aTFarray )
+    {
+    	fTFarray = aTFarray;
+    }
+    std::vector<std::complex<double>> AnalyticResponseFunction::GetTFarray()
+    {
+    	return fTFarray;
+    }
+
+
+
+
+
 
 
 
--- a/Source/RxComponents/LMCAnalyticResponseFunction.cc
+++ b/Source/RxComponents/LMCAnalyticResponseFunction.cc
@@ -52,6 +52,14 @@ namespace locust
     {
     	return fTFarray;
     }
+    void AnalyticResponseFunction::SetGFarray( std::vector<std::pair<double,double>> aGFarray )
+    {
+    	fGFarray = aGFarray;
+    }
+    std::vector<std::pair<double,double>> AnalyticResponseFunction::GetGFarray()
+    {
+    	return fGFarray;
+    }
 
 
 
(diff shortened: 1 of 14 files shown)


Source/RxComponents/LMCAnalyticResponseFunction.cc
```diff
@@ -0,0 +1,64 @@
...
+	    fGeneratingTF( false ),
+		fInitialFreq( 0. ),
+		fTFarray(0 )
+    {}
+    AnalyticResponseFunction::~AnalyticResponseFunction() {}
+
+
+    bool AnalyticResponseFunction::Configure( const scarab::param_node& aParam )
+    {
+        return true;
```
marsteller, line 28: Is this just a placeholder for the time being and will be filled in later, or is there really nothing meaningful to configure here from the `aParam`?

Source/RxComponents/LMCDampedHarmonicOscillator.hh
```diff
@@ -0,0 +1,52 @@
...
+
+
+        private:
+            double fCavityFrequency; // Hz
+            double fCavityOmega; // radians/s
+            double fCavityQ;
+            double fCavityDampingFactor;
+            double fBFactor; // harmonic oscillator parameter
+            double fCavityOmegaPrime;  // damped resonant frequency
+            int fMaxNBins;
```
marsteller, line 39: Is there a reason that we need an outside defined `fMaxNBins`? If not, I think that calculating the number of bins from the `fCavityQ` and `fTimeResolution` until the response has decayed below the `fThresholdFactor` would be more straigthforward.

Source/RxComponents/LMCDampedHarmonicOscillator.cc
```diff
@@ -0,0 +1,95 @@
...
+{
+
+	LOGGER( lmclog, "DampedHarmonicOscillator" );
+
+    DampedHarmonicOscillator::DampedHarmonicOscillator():
+    		fMaxNBins( 4000 ),
+			fTimeResolution( 1.e-12 ),
+			fCavityFrequency( 1.0e9 ),
+			fCavityQ( 1000 ),
+			fThresholdFactor ( 0.25 ),
```
marsteller, line 21: The combination of preset value for `fMaxNBins` doesn't make much sense given the other preset values. For the given parameters, `fMaxNBins` needs to be at least ~4.4e5 to contain everything up the `fThresholdFactor` of `0.25`. The current number of `4000` bins will only go down to `0.9875` of the initial value. 
Also, setting the threshold to 1/4th sounds like it would affect the resulting output by quite a bit.

The need to have such a long impulse response is something that has been concerning me for a while though ☹️

marsteller (approved):
As far as I can understand the changes, they look good to me. I commented on the `fMaxNBins` for the oscillator, but that's more of a general discussion point.

Source/RxComponents/LMCAnalyticResponseFunction.cc
pslocum: Right, it is a placeholder to be filled in as needed.  At the moment we are doing the configuration in its derived classes DampedHarmonicOscillator and EquivalentCircuit.

Source/RxComponents/LMCDampedHarmonicOscillator.hh
pslocum: I was hoping to use `fMaxNBins` as an upper limit to avoid configuring something that takes longer than we anticipated.  Also, we have a debugging file that outputs the configured FIR in a text file for inspection, here:   https://github.com/project8/locust_mc/blob/master/Source/Core/LMCHFSSResponseFileHandler.cc#L135 .

Source/RxComponents/LMCDampedHarmonicOscillator.cc
pslocum: Aha thanks, we will be using the debugging FIR.txt file to check this type of issue.  But yes, they are arbitrary values at the moment.

Source/RxComponents/LMCDampedHarmonicOscillator.hh
pslocum: In principle we could exit the code with an error/warning if fMaxBins is required.

(Almost ready to test.)

pslocum:
The default values in DampedHarmonicOscillator are producing this FIR.  It needs optimization (it still looks too short).
discreteGF

marsteller:
The number of bins you need to reach the threshold is just given by the exponential decay part of the Green's function. You can plug in the cavity parameters and solve for the time needed for that exponential to decay to the threshold. Then the number of bins is given by dividing through the time resolution.  

Since this is the physics part of the problem, I don't think you can optimize this in any way to get less bins unless you make the time resolution larger.

pslocum:
With the DampedHarmonicOscillator (DHO) in place, we have a compact sideband pattern for pitch=89.5 degrees in the large CDR trap.  These lines will likely change as mode coupling gets refined in the next steps.  With the present changes:

* new analytic response function DHO implemented and tuned as in suggestions by @marsteller .
* new class structure to make the DHO selected by default, but replaceable with command line parameter.
* new flags for writing intermediate files.  

I am thinking to merge to develop so that the next steps can be discussed and addressed in a new pull request.  Does this sound OK?
sidebands

PR #243, opened 2022-08-08
Feature/cavity development

pslocum:
These changes include a possible placeholder for a cavity power model in LMCDampedHarmonicOscillator.  Also, the wave impedance is reverted back to the expression in Jackson, in cgs->si units.  The goal is to allow for more stringent tests of the normalization while maintaining modularity.

### DIFF
--- a/Source/Fields/LMCCylindricalCavity.cc
+++ b/Source/Fields/LMCCylindricalCavity.cc
@@ -94,58 +94,48 @@ namespace locust
 
     double CylindricalCavity::Z_TE(int l, int m, int n, double fcyc) const
     {
+    	/*
     	double Z_TE = 1.0;
     	double x_lm = fInterface->fBesselNKPrimeZeros[l][m];
     	double k1 = x_lm / GetDimR();
     	double k3 = n * LMCConst::Pi() / GetDimL();
     	double k = pow(k1*k1+k3*k3,0.5);
     	double k0 = fcyc / LMCConst::C();
-    	double A = 1.;
-    	double B = k*k;
-    	double C = k0*k0;
-    	double Q = 1.;  // Q for suppressed modes.
-    	if ((l==0)&&(m==1)&&(n==1)) Q = 1000.; // Q for lmn = 011.
-    	double denom = (Q*B - Q*C - C) * (Q*B - Q*C - C) + C*C;
-    	double real = (Q*A * (Q*B - Q*C - C)) / denom;
-    	double imag = -Q*A*C / denom;
 
     	if ( k*k-k0*k0 != 0. )
     	{
-    		// Red Jackson waveguide impedance Eq. 8.32
+    		// Red Jackson wave impedance Eq. 8.32
     		Z_TE = LMCConst::MuNull() * fcyc / LMCConst::C() / k;
-    		// after Collin Foundations of M.E. Eq. 7.132
-//    		Z_TE *= fcyc * LMCConst::MuNull() / (k0*k0 - k*k);  // neglect Q term.
-//    		Z_TE *= fcyc * LMCConst::MuNull() * sqrt( real*real + imag*imag ); // mag
     	}
+    	*/
+
+    	// Bypass all realistic impedance calculations for now, and return a constant.
+    	// This allows us to implement a power model elsewhere in the code without having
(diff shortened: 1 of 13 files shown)


marsteller:
I think this looks good.

Hamish and I are currently discussing the details of the Hanneke coupling factor and how to implement that in the impulse picture. Since the formula contains the radius and axial position, as well as the cyclotron frequency I don't know how straight forward it will be.

pslocum:
Thanks.  Yes, Jeremy and I were also noticing that the Hanneke coupling
factor contains the radius and axial position.  We were under the
impression that it represents the mode amplitude at the location of the
electron.  If that is the case, then we might be ok because the mode
amplitude at the electron (in r and z) is already present as a factor,
here:
https://github.com/project8/locust_mc/blob/feature/cavityDevelopment/Source/Generators/LMCCavitySignalGenerator.cc#L597
.  As for the cyclotron frequency, we are presently baking it in when
convolving the Kass current impulses with the damped harmonic oscillator
Green's function, here:
https://github.com/project8/locust_mc/blob/feature/cavityDevelopment/Source/Kassiopeia/LMCFieldCalculator.cc#L168
.

We can of course double check everything in the meantime.

Source/Generators/LMCCavitySignalGenerator.cc
```diff
@@ -579,29 +574,29 @@ namespace locust
...
 
-    					if (!fE_Gun)
+    					if (!fInterface->fE_Gun)
     					{
-    						unitConversion = 1.;  // mks units in Collin amplitudes.
-    						excitationAmplitude = fAvgDotProductFactor * modeAmplitude * cavityFIRSample;
+    						// sqrt(4PIeps0) for Kass current si->cgs, sqrt(4PIeps0) for A_lambda coefficient cgs->si
+    						unitConversion = 1. / LMCConst::FourPiEps(); // see comment
+    						excitationAmplitude = fAvgDotProductFactor * modeAmplitude * cavityFIRSample * fInterface->fField->Z_TE(l,m,n,tKassParticleXP[7]) * 2. * LMCConst::Pi() / LMCConst::C() / 1.e2;
     						std::vector<double> tProbeLocation = {fInterface->fField->GetDimR()*fPowerCombiner->GetCavityProbeRFrac(), 0., fPowerCombiner->GetCavityProbeZ()};
```
jkgaison65, line 591: I know this push has some changes between cgs and SI units and I always get confused in those differences, but at least at first glance there apparently seems to be different units here vs the fE_gun "else" block definition of the `excitationAmplitude` variable. Everything is identical besides the Z_TE (in ohms?) and the ScaleEPoyntingVector (in sqrt of power flux?) here. I don't think this really matters at this stage due to the arbitrary wave impedance value anyway, but it's worth keeping in mind.

Source/Fields/LMCCylindricalCavity.cc
```diff
@@ -94,54 +94,48 @@ namespace locust
...
+    		// Red Jackson wave impedance Eq. 8.32
+    		Z_TE = LMCConst::MuNull() * fcyc / LMCConst::C() / k;
     	}
+    	*/
+
+    	// Bypass all realistic impedance calculations for now, and return a constant.
+    	// This allows us to implement a power model elsewhere in the code without having
+    	// a duplicate resonant impedance from this function.
+    	double Z_TE = 500.; // ohms
     	return Z_TE;
```
jkgaison65, line 116: Just to confirm this is an arbitrary constant at this stage, right? Is there any reason this is set at 500 rather than just 1?

jkgaison65 (approved):
The placeholder Hanneke power factor looks like a clean, mode-independent scale that can be worked with. I've added a couple minor comments but overall it looks solid to me.

Source/Generators/LMCCavitySignalGenerator.cc
pslocum: There might be some more ways to check this, such as not calculating a Poynting vector and instead driving the probe with the normalized E-field, and then multiplying the Katydid output by the cross-secional area of the waveguide.  Since we moved to a complex FIR there is still some normalization that needs a closer look.

pslocum:
Suggesting to merge these changes now so that they can be used in the next steps.  Thanks for the comments.

Issue #245, opened 2022-10-17
Precision in Apptainer on RedHat 7 is different than in Docker on Ubuntu


pslocum:
Locust-Kass time series outputs are different in the two cases for the cavity simulation.  The problem improves if double precision variables are converted to long double precision, but the output is still not identical.  Depending on impact, it might be necessary to consider compiling with long double precision fftw libraries.

pslocum:
This issue appears to be due to incomplete separation between the signal and feedback calculations.  On a laptop the problem has not been noticeable, but in the cluster environment there small but noticeable effects.  The next step toward the complete separation will be to configure the Kass feedback calculations using `Configure( const scarab::param_node& aParam )` , as is presently done elsewhere in Locust.

pslocum:
This issue has been traced to the issue described in https://github.com/project8/locust_mc/issues/246 .

PR #247, opened 2022-10-31
Feature/cavity development

pslocum:
The changes here allow for more stringent separation in memory between cavity excitations in Locust and Kassiopeia.  At the same time, the nominal configuration parameters are now readable on both sides of the Kass-Locust interface.  This allows for a single configuration file but two separate calculations in Locust and Kass.


pslocum:
This addresses part of the issue described in https://github.com/project8/locust_mc/issues/245 .

pslocum:
Merging into develop.  The next steps will be to increase visibility of Locust and Kass libraries in unit tests.

PR #257, opened 2023-01-05
Feature/build kass3.8.2

pslocum:
These changes include progress toward building Kass v3.8.2 as a submodule, using the new PkgConfig *.pc files.


pslocum:
Additional optimization of the Locust-Kass interface will be needed for configuring updated physical constants.   And, VTK is now supported in the Locust-Kass build:
image

PR #273, opened 2023-04-21
Feature/polarization integration

jkgaison65:
New version of alternate mode implementation which better fits restructuring of Field classes for cylindrical cavities. Defaults to using TE011 mode, but expanded to implement higher order l>0 modes for TE_lmn with arbitrary polarizations based on electron location.


pslocum (commented):
These expansions look fantastic, in particular with the generalization to other modes.  I have made some comments based on recent restructuring, and will do a draft pull request with suggested details.

PR #277, opened 2023-05-03
feature/scatteringChecks

pslocum:
This draft pull request has been created to test scattering in Kassiopeia.


pslocum:
This work will likely be paused pending final checks on the cluster.  With signal calculations compatible with Kassiopeia scattering interactions, the likely next step for future work will be to define appropriate event terminators in the xml config file.

PR #279, opened 2023-05-15
Feature/multi probe

pslocum:
This branch contains changes toward having two independently configured probes in the cavity simulation, each with its own DAQ channel.  More testing is needed before merging.


jkgaison65:
Overall looks like a clean implementation of a second channel. Noise is generated independently for each channel when that generator is run as well. Right now probes are identical up to a gain factor. In the future it's probably worth implementing a way to differentiate probes more specifically (i.e. make a probe couple to a different field than just E_theta, or allow a probe to be tuned to a different frequency), but this is a great start.

pslocum:
Thanks for the comments, and agreed on future refinements as needed.

PR #281, opened 2023-06-07
Feature/multi mode

pslocum:
These changes continue to consolidate field-related calculations into designated classes, allowing for easier access.  This is a draft pull request, pending more testing.

### DIFF
--- a/Source/Kassiopeia/LMCFieldCalculator.hh
+++ b/Source/Kassiopeia/LMCFieldCalculator.hh
@@ -40,10 +40,10 @@ namespace locust
             double GetDampingFactorCavity(Kassiopeia::KSParticle& aFinalParticle);
             double GetCouplingFactorTM01(Kassiopeia::KSParticle& aFinalParticle);
             double GetCouplingFactorTE10(Kassiopeia::KSParticle& aFinalParticle);
-            double GetCouplingFactorTElmnCavity(int l, int m, int n, bool bTE, Kassiopeia::KSParticle& aFinalParticle);
+            double GetCouplingFactorTXlmnCavity(int l, int m, int n, bool bTE, Kassiopeia::KSParticle& aFinalParticle);
             double GetTM01FieldWithTerminator(Kassiopeia::KSParticle& aFinalParticle);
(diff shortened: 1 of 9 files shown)


Source/Kassiopeia/LMCFieldCalculator.hh
jkgaison65, line 1: Overall this looks like a nice clean implementation expanding to TM modes. My only comment is that the name for this function implies that it is specific for TE modes while the input allows you to select between TE and TM so I'd remove 'TE' from the function name

Source/Kassiopeia/LMCFieldCalculator.hh
```diff
@@ -32,17 +32,18 @@ namespace locust
...
             double GetDampingFactorPhase1(Kassiopeia::KSParticle& aFinalParticle);
             double GetDampingFactorCavity(Kassiopeia::KSParticle& aFinalParticle);
             double GetCouplingFactorTM01(Kassiopeia::KSParticle& aFinalParticle);
             double GetCouplingFactorTE10(Kassiopeia::KSParticle& aFinalParticle);
-            double GetCouplingFactorTE011Cavity(Kassiopeia::KSParticle& aFinalParticle);
+            double GetCouplingFactorTElmnCavity(int l, int m, int n, bool bTE, Kassiopeia::KSParticle& aFinalParticle);
             double GetTM01FieldWithTerminator(Kassiopeia::KSParticle& aFinalParticle);
             double GetTE10FieldAfterOneBounce(Kassiopeia::KSParticle& aFinalParticle);
-            double GetTE011FieldCavity(Kassiopeia::KSParticle& aFinalParticle);
+            double GetTElmnFieldCavity(int l, int m, int n, Kassiopeia::KSParticle& aFinalParticle);
```
jkgaison65, line 46: Similarly we'll want to generalize this to both TE and TM modes soon, so it's probably worth removing the 'TE' from this function name and add a 'bool bTE' input here too, even if it isn't currently utilized in the actual function definition

jkgaison65 (approved):
Changes look great! Just a couple minor comments in the FieldCalculator but other than that I think it's good to merge

PR #282, opened 2023-06-14
implemented TM polarization into recent multiMode work, includes teMo…

jkgaison65:
Adds in polarization of TM modes as well as H fields. Incorporates configurable teMode bool that selects whether the TE modes are used (the default) or if TM modes are used in the DriveMode() function.

### DIFF
--- a/Source/Fields/LMCCylindricalCavity.cc
+++ b/Source/Fields/LMCCylindricalCavity.cc
@@ -309,7 +309,7 @@ namespace locust
                 //If mode has theta dependence, mode polarization is set by electron location. Probe coupling must be set relative to that angle
                 double thetaElectron = tKassParticleXP[1];
                 double TMOffset = 0;
-                if(!teMode) TMOffset = LMCConst::Pi() / 2.0 / (double)l; //Different angular dependence of E_theta [sin vs cos] in TM modes requires offset to correct for probe projection 
+                //if(!teMode) TMOffset = LMCConst::Pi() / 2.0 / (double)l; //Different angular dependence of E_theta [sin vs cos] in TM modes requires offset to correct for probe projection 
                 thetaEffective.push_back(thetaProbe[0] - thetaElectron + TMOffset);
                 thetaEffective.push_back(thetaProbe[1] - thetaElectron + TMOffset);
     	}
--- a/Source/Fields/LMCCylindricalCavity.cc
+++ b/Source/Fields/LMCCylindricalCavity.cc
@@ -335,34 +335,34 @@ namespace locust
     	double tTheta = tKassParticleXP[1];
     	double tZ = tKassParticleXP[2];
        	std::vector<double> tField;
-	double normFactor;
-
-	if(teMode)
-	{
+       	double normFactor;
+       	if(teMode)
+       	{
        		tField = fFieldCore->TE_E(GetDimR(),GetDimL(),l,m,n,tR,tTheta,tZ,includeOtherPols);
        		normFactor = GetNormFactorsTE()[l][m][n];
-	}
-	else
-	{
-                tField = fFieldCore->TM_E(GetDimR(),GetDimL(),l,m,n,tR,tTheta,tZ,includeOtherPols);
-                normFactor = GetNormFactorsTM()[l][m][n];
-	}
-   		auto it = tField.begin();
-   		while (it != tField.end())
-   		{
-   			if (!isnan(*it))
-   			{
-   				(*it) *= normFactor;
-   			}
-   			else
-   			{
-   				(*it) = 0.;
-   			}
-   			*it++;
-   		}
+       	}
+       	else
+       	{
+       		tField = fFieldCore->TM_E(GetDimR(),GetDimL(),l,m,n,tR,tTheta,tZ,includeOtherPols);
+       		normFactor = GetNormFactorsTM()[l][m][n];
+       	}
+       	auto it = tField.begin();
+       	while (it != tField.end())
+       	{
+       		if (!isnan(*it))
+       		{
+       			(*it) *= normFactor;
+       		}
+       		else
+       		{
+       			(*it) = 0.;
+       		}
+       		*it++;
+       	}
 
        	return tField;  // return normalized field.
-       }
+    }
+
 
 	double CylindricalCavity::TotalFieldNorm(std::vector<double> field)
 	{
--- a/Source/Fields/LMCCylindricalCavity.cc
+++ b/Source/Fields/LMCCylindricalCavity.cc
@@ -331,11 +331,6 @@ namespace locust
 
     }
 
-    std::vector<double> CylindricalCavity::GetFieldAtProbe(int l, int m, int n, bool includeOtherPols, std::vector<double> tKassParticleXP)
-    {
-	return GetFieldAtProbe(l, m, n, includeOtherPols, tKassParticleXP, true);
-    }
-
     std::vector<double> CylindricalCavity::GetNormalizedModeField(int l, int m, int n, std::vector<double> tKassParticleXP, bool includeOtherPols, bool teMode)
     {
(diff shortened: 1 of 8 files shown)


Source/Fields/LMCCylindricalCavity.hh
```diff
@@ -56,6 +57,7 @@ namespace locust
             virtual void CheckNormalization(int nModes);
             virtual void PrintModeMaps(int nModes, bool bTE, double zSlice);
             virtual std::vector<double> GetFieldAtProbe(int l, int m, int n, bool includeOtherPols, std::vector<double> tKassParticleXP);
+            virtual std::vector<double> GetFieldAtProbe(int l, int m, int n, bool includeOtherPols, std::vector<double> tKassParticleXP, bool teMode);
```
pslocum, line 60: Can these two GetFieldAtProbe() functions be consolidated into just one function, or do they need to be overloaded?

Source/Fields/LMCCylindricalCavity.hh
```diff
@@ -48,6 +48,7 @@ namespace locust
             virtual double Integrate(int l, int m, int n, bool teMode, bool eField);
             virtual std::vector<double> GetDopplerFrequency(int l, int m, int n, std::vector<double> tKassParticleXP);
             virtual std::vector<double> GetNormalizedModeField(int l, int m, int n, std::vector<double> tKassParticleXP, bool includeOtherPols);
+	    virtual std::vector<double> GetNormalizedModeField(int l, int m, int n, std::vector<double> tKassParticleXP, bool includeOtherPols, bool teMode);
```
pslocum, line 51: Can these two GetNormalizedModeField() functions be consolidated into one function, or do we need both of them?

Source/Fields/LMCField.hh
```diff
@@ -72,6 +72,7 @@ namespace locust
             virtual std::vector<double> GetDopplerFrequency(int l, int m, int n, std::vector<double> tKassParticleXP) {return {0.};};
             virtual std::vector<double> GetNormalizedModeField(int l, int m, int n, std::vector<double> tKassParticleXP) {return {0.};};
             virtual std::vector<double> GetNormalizedModeField(int l, int m, int n, std::vector<double> tKassParticleXP, bool includeOtherPols) {return {0.};};
+	    virtual std::vector<double> GetNormalizedModeField(int l, int m, int n, std::vector<double> tKassParticleXP, bool includeOtherPolsi, bool teMode) {return {0.};};
```
pslocum, line 75: Similarly, can these two GetNormalizedModeField() functions be combined into just one function?

Source/Fields/LMCField.hh
```diff
@@ -80,6 +81,7 @@ namespace locust
             virtual void CheckNormalization(int nModes){};
             virtual void PrintModeMaps(int nModes, bool bTE, double zSlice){};
             virtual std::vector<double> GetFieldAtProbe(int l, int m, int n, bool includeOtherPols, std::vector<double> tKassParticleXP){return {0.};};
+            virtual std::vector<double> GetFieldAtProbe(int l, int m, int n, bool includeOtherPols, std::vector<double> tKassParticleXP, bool teMode){return {0.};};
```
pslocum, line 84: Combine GetFieldAtProbe() functions?

Source/Fields/LMCCylindricalCavity.cc
```diff
@@ -322,22 +324,36 @@ namespace locust
 
     	//Assumes probe couples to E_theta of mode. If mode is polarized, transforms angle to reference frame of electron
     	std::vector<double> tEFieldAtProbe;
-    	tEFieldAtProbe.push_back(GetNormalizedModeField(l,m,n,tProbeLocation[0],0).back());
-    	tEFieldAtProbe.push_back(GetNormalizedModeField(l,m,n,tProbeLocation[1],0).back());
+    	tEFieldAtProbe.push_back(GetNormalizedModeField(l,m,n,tProbeLocation[0],0,teMode).back());
+    	tEFieldAtProbe.push_back(GetNormalizedModeField(l,m,n,tProbeLocation[1],0,teMode).back());
```
pslocum, line 328: Are we missing some of the TM111 mode here?  It looks like we are only allowing the probe to read out the theta component.  That should work for the TE011 mode, but does it give us the total TM111 mode power?    I am actually wondering if we will need to rethink the probe.  Thoughts?

pslocum (commented):
Thanks for doing this.  It looks great.  I made a few comments in the request.  Looking forward to hearing your thoughts.

Source/Fields/LMCCylindricalCavity.cc
jkgaison65: I guess the question is do we want to be able to use Locust to get the total mode power at a given location or do we want to model the response of an actual readout probe. A probe may only be able to couple to some field component (e.g. a wire loop readout would only measure the H field normal to the loop).

Source/Fields/LMCCylindricalCavity.hh
jkgaison65: I implemented it as overloaded functions originally in case there were some calls to the original versions in other parts of the code and I didn't want to break other functionality, but after some quick tests it looks like it's fine with only the newest version of GetFieldAtProbe() and GetNormalizedModeField(). I've pushed the changes for both functions to remove the extra definitions

Source/Fields/LMCCylindricalCavity.cc
jkgaison65: I've updated both the GetFieldAtProbe() function and the modeAmplitude variable to use the norm of the entire electric field. This remains identical for the TE011 mode (which only has an Etheta component anyway), and while the TM111 mode changes in amplitude its azimuthal polarization remains qualitatively the same as before with respect to the readout probe.

Source/Fields/LMCCylindricalCavity.hh
pslocum: Thanks for checking all of that.

Source/Fields/LMCCylindricalCavity.cc
```diff
@@ -355,6 +364,18 @@ namespace locust
...
+	{
+		double norm = 0;
+		auto it = field.begin();
+		while (it != field.end())
+		{
+			if (!isnan(*it)) norm += (*it)*(*it);
+			*it++;
+		}
+		return sqrt(norm);
+	}
```
pslocum, line 377: Thanks, this looks great.

Source/Generators/LMCCavitySignalGenerator.cc
```diff
@@ -352,11 +358,11 @@ namespace locust
     				if (fFieldCalculator->ModeSelect(l, m, n, fInterface->fE_Gun, fNormCheck))
     				{
     					std::vector<double> tE_normalized;
-    					tE_normalized = fInterface->fField->GetNormalizedModeField(l,m,n,tKassParticleXP,1);
+    					tE_normalized = fInterface->fField->GetNormalizedModeField(l,m,n,tKassParticleXP,1,fbTE);
     					double cavityFIRSample = fFieldCalculator->GetCavityFIRSample(tKassParticleXP, fBypassTF).first;
     					dopplerFrequency = fInterface->fField->GetDopplerFrequency(l, m, n, tKassParticleXP);
     					double tAvgDotProductFactor = fInterface->fField->CalculateDotProductFactor(l, m, n, tKassParticleXP, tE_normalized, tThisEventNSamples);
-    					double modeAmplitude = sqrt( tE_normalized.front()*tE_normalized.front()  + tE_normalized.back()*tE_normalized.back());
+    					double modeAmplitude = fInterface->fField->TotalFieldNorm(tE_normalized);
```
pslocum, line 365: Nice.

pslocum (commented):
These changes look really great.  If we have time, I am going to try a couple of possible next steps and do a pull request into this branch, with a request for your feedback.

PR #283, opened 2023-06-19
Feature/option multi mode

pslocum:
This is a draft pull request for discussion/disposal/etc.  The idea is to use a boolean flag to select or reject a multimode running configuration using superposition, where individual modes could possibly be read out using parametrized positioning of two probes.  This running mode may or may not align with near term goals, such as running with one uniquely select mode at a time.  It also does not calculate mode mixing.


jkgaison65:
Overall this looks clean and is a welcome restructuring of the TM/multimode enabling parameter!

jkgaison65 (approved):
Looks great, and it's a clean reorganization and generalization of the TM/Multimode parameter

PR #284, opened 2023-06-21
Feature/option multi mode

pslocum:
These changes allow the phases for each mode to evolve separately at the signal calculation.


Source/RxComponents/LMCCavityModes.cc
```diff
@@ -61,15 +61,31 @@ namespace locust
             }
         }
 
+
+        fVoltagePhase.resize(2); // max nchannels.  TO-DO:  Configure this as LMCGenerator::fNChannels.
```
jkgaison65, line 65: It looks like the resizing doesn't line up with the indexing here in general. You start by resizing fVoltagePhase for 2 channels, but then iterate over nModes. Here they happen to both be 2, but I think the first for loop should index over nchannels, the second and third loops should index over GetNCavityModes(), and then you won't need a third GetNCavityModes() loop. Still 3 for loops total though.

Source/RxComponents/LMCCavityModes.cc
```diff
@@ -61,15 +61,31 @@ namespace locust
...
 
-	bool CavityModes::AddOneModeToCavityProbe(Signal* aSignal, std::vector<double> particleXP, double excitationAmplitude, double EFieldAtProbe, std::vector<double> cavityDopplerFrequency, double dt, double phi_LO, double totalScalingFactor, unsigned sampleIndex, int channelIndex, bool initParticle)
+	bool CavityModes::AddOneModeToCavityProbe(int l, int m, int n, Signal* aSignal, std::vector<double> particleXP, double excitationAmplitude, double EFieldAtProbe, std::vector<double> cavityDopplerFrequency, double dt, double phi_LO, double totalScalingFactor, unsigned sampleIndex, int channelIndex, bool initParticle)
 	{
 		double dopplerFrequency = cavityDopplerFrequency[0];  // Only one shift, unlike in waveguide.
-		SetVoltagePhase( GetVoltagePhase(channelIndex) + dopplerFrequency * dt, channelIndex ) ;
+		SetVoltagePhase( GetVoltagePhase(channelIndex, l, m, n)[channelIndex][l][m][n] + dopplerFrequency * dt, channelIndex, l, m, n ) ;
 		double voltageValue = excitationAmplitude * EFieldAtProbe;
-		voltageValue *= cos(GetVoltagePhase(channelIndex));
+		voltageValue *= cos(GetVoltagePhase(channelIndex, l, m, n)[channelIndex][l][m][n]);
```
jkgaison65, line 88: You've defined GetVoltagePhase as a 4D vector, so you don't need the parameters in the function call (or alternatively you could actually use the parameters in the function and return a scalar double and then not use the indices here).

Source/RxComponents/LMCCavityModes.cc
```diff
@@ -164,14 +180,14 @@ namespace locust
 		return true;
 	}
 
-    double CavityModes::GetVoltagePhase(unsigned aChannel)
+    std::vector<std::vector<std::vector<std::vector<double>>>> CavityModes::GetVoltagePhase(int aChannel, int l, int m, int n)
```
jkgaison65, line 183: Same comment as when this is called. Personally, I'd vote for using the aChannel, l, m, and n parameters within this definition and just return a scalar. This would match the structure of the SetVoltagePhase() definition below

Source/RxComponents/LMCCavityModes.hh
```diff
@@ -36,17 +36,17 @@ namespace locust
             CavityModes();
             virtual ~CavityModes();
             virtual bool Configure( const scarab::param_node& aNode );
-        	virtual bool AddOneModeToCavityProbe(Signal* aSignal, std::vector<double> particleXP, double excitationAmplitude, double EFieldAtProbe, std::vector<double> dopplerFrequency, double dt, double phi_LO, double totalScalingFactor, unsigned sampleIndex, int channelIndex, bool initParticle);
+        	virtual bool AddOneModeToCavityProbe(int l, int m, int n, Signal* aSignal, std::vector<double> particleXP, double excitationAmplitude, double EFieldAtProbe, std::vector<double> dopplerFrequency, double dt, double phi_LO, double totalScalingFactor, unsigned sampleIndex, int channelIndex, bool initParticle);
         	virtual bool AddOneSampleToRollingAvg(int l, int m, int n, double excitationAmplitude, unsigned sampleIndex);
-            double GetVoltagePhase(unsigned aChannel);
-            void SetVoltagePhase( double aPhase, unsigned aChannel );
+            std::vector<std::vector<std::vector<std::vector<double>>>> GetVoltagePhase(int aChannel, int l, int m, int n);
```
jkgaison65, line 41: If you take the suggestion above, this will just be a scalar double.

jkgaison65 (changes requested):
Right now the indexing happens to work out since nchannels and nmodes both happen to be 2, but in general won't scale. Also there are some redundant uses of the indices than can be condensed.

pslocum:
Thanks for the great comments.  We could maybe either move this pull request into another branch for revisions, apply the changes in this branch, or possibly close this request and do some combination of both of those things.

pslocum:
Thank you for finding the overuse of the indices.  The code looks better now.  I will plan to run this branch on the cluster to confirm that the mode superposition calculations are unchanged, and then merge to develop.

pslocum:
Cluster testing shows nominal looking outputs given previous checks of mode superpositions.

PR #295, opened 2023-09-19
Feature/MultiMode

jkgaison65:
Updates DHO model to accommodate independent configurations for different modes


pslocum (commented):
These are exciting updates, thank you!  I have added a few comments.  Mainly I am wondering if we should replace the existing functions with your multimode upgrade, since it is more general.  (e.g. replace fFilterComplex with fFilterComplexArray.)  I can help with this too.  Will follow up offline for more thoughts.  Thanks again!

jkgaison65:
Expansion of MultiMode features to a broader range of parameters, such as DHO Threshold and frequency (usually not needed, but may be needed if we ever look at higher harmonics). Also generalized configurable parameters beyond just TE011 and TM111 modes

PR #300, opened 2023-11-13
Import mode maps

pslocum:
The purpose of this draft feature branch is to test mode map uploads, interpolation, and normalization.


jkgaison65:
Update that allows for interpolation of an input mode map. Currently it's set to only use the TE011 mode, further expansion will be needed once actual mode maps exist for use. Double check the Eigen3 library is working (e.g. the Dense library called in Source/Fields/LMCModeMapCavity.cc), as I got that path working locally but didn't thoroughly check paths on machines besides my own.

pslocum:
This is amazing progress.  I have pushed a test to find the Eigen3 package with cmake, for use in preliminary testing in the container environment.

PR #304, opened 2023-12-05
Hotfix/rpath fix

nsoblath:


pslocum (commented):
Hi @nsoblath , thanks very much for these changes.  I am wondering whether we could pull this into develop instead of master, or whether there is a reason not to do that?

PR #321, opened 2024-10-29
Feature/versionUpdates

pslocum:
These changes include an update to the recent release of Kassiopeia, as well as some minor additional configurability for multi-probe cavity readouts.

### DIFF
--- a/Source/RxComponents/LMCCavityModes.hh
+++ b/Source/RxComponents/LMCCavityModes.hh
@@ -49,7 +49,7 @@ namespace locust
 
         private:
             std::vector<std::vector<std::vector<std::vector<double>>>> fVoltagePhase;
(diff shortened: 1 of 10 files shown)


Source/RxComponents/LMCCavityModes.hh
```diff
@@ -49,6 +49,7 @@ namespace locust
 
         private:
             std::vector<std::vector<std::vector<std::vector<double>>>> fVoltagePhase;
+            std::vector<double> fChannelPhaseOffsetDeg;
```
jkgaison65, line 52: This variable is labeled as "Deg" in the name but is actually in radians. This could be confusing, and I'd recommend dropping "Deg" from the name

PR #325, opened 2025-03-20
updates to randomization and metadata format

pslocum:
This pull request includes two changes:  If a random seed is not configured, the internal random seed is now generated in microseconds instead of seconds.  This allows for arrays of hpc jobs to be randomized independently.


pslocum:
The second change includes updates to the metadata format such that floats are no longer formatted as strings.

